PR #8, opened 2014-06-10
Allow for query parameter on has_many :videos

stavro:
@claudiofullscreen Would love to get your thoughts.

I'm trying to replace youtube_it with YT, but currently the only missing piece is querying for videos.

This allows for the following usage (tested on my account):

``` ruby
YT::Account.new(credentials).videos.map &:title
 => ["asdf",  "Vegas Indoor Skydiving", "Swing", "UCSD Dance by the Shores Adv. Hustle"]

YT::Account.new(credentials).videos(q: 'vegas').map &:title
=> ["Vegas Indoor Skydiving"]
```

Wanted to get your thoughts before i spend time fixing specs.

### DIFF
--- a/lib/yt/collections/videos.rb
+++ b/lib/yt/collections/videos.rb
@@ -5,6 +5,11 @@ module Yt
   module Collections
     class Videos < Base
 
+      def initialize(options)
+        @q = options[:q]
+        super
+      end
+
     private
 
       def new_item(data)
@@ -13,7 +18,14 @@ module Yt
 
       def list_params
         super.tap do |params|
-          params[:params] = {channelId: @parent.id, type: :video, maxResults: 50, part: 'snippet'}
+          params[:params] = {type: :video, maxResults: 50, part: 'snippet'}
+
+          if @parent.auth == @auth
+            params[:params].merge! forMine: true, q: @q
+          else
+            params[:params].merge! channelId: @parent.id
+          end
+
           params[:path] = '/youtube/v3/search'
         end
       end
--- a/lib/yt/collections/base.rb
+++ b/lib/yt/collections/base.rb
@@ -15,8 +15,8 @@ module Yt
         @auth = options[:auth]
       end
 
-      def self.of(parent)
-        new parent: parent, auth: parent.auth
+      def self.of(parent, options = {})
+        new options.merge parent: parent, auth: parent.auth
       end
     end
   end
--- a/lib/yt/models/base.rb
+++ b/lib/yt/models/base.rb
@@ -18,9 +18,9 @@ module Yt
         mod = attributes.sub(/.*\./, '').camelize
         collection = "Yt::Collections::#{mod.pluralize}".constantize
 
-        define_method attributes do
+        define_method attributes do |options = {}|
           ivar = instance_variable_get "@#{attributes}"
-          instance_variable_set "@#{attributes}", ivar || collection.of(self)
+          instance_variable_set "@#{attributes}", ivar || collection.of(self, options)
         end
       end
 


lib/yt/collections/videos.rb
```diff
@@ -13,7 +18,14 @@ def new_item(data)
 
       def list_params
         super.tap do |params|
-          params[:params] = {channelId: @parent.id, type: :video, maxResults: 50, part: 'snippet'}
+          params[:params] = {type: :video, maxResults: 50, part: 'snippet'}
+
+          if @parent.auth == @auth
+            params[:params].merge! forMine: true, q: @q
```
stavro, line 24: the `q` parameter doesn't work if you also specify a `channelId`... why? No idea. this isn't specified in the docs

claudiofullscreen, line 24: I'll look at it now

On Tue, Jun 10, 2014 at 2:52 PM, Sean Stavropoulos <notifications@github.com

> wrote:
> 
> In lib/yt/collections/videos.rb:
> 
> > @@ -13,7 +18,14 @@ def new_item(data)
> > 
> > ```
> >    def list_params
> >      super.tap do |params|
> > ```
> > -          params[:params] = {channelId: @parent.id, type: :video, maxResults: 50, part: 'snippet'}
> > -          params[:params] = {type: :video, maxResults: 50, part: 'snippet'}
> >   +
> > -          if @parent.auth == @auth
> > -            params[:params].merge! forMine: true, q: @q
> 
> the q parameter doesn't work if you also specify a channelId... why? No
> idea. this isn't specified in the docs
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/Fullscreen/yt/pull/8/files#r13622141.

## 

 Claudio Baccigalupo  | Software Writer  |  Fullscreen, Inc.
http://Fullscreen.net  |  +1.310.202.3333

claudiofullscreen, line 24: it's a good idea, but let me figure out this conflict between ?q and ?channelId

claudiofullscreen, line 24: @stavro Ok I see the problem. Adding `forMine: true` here is not the solution, because you would limit `channel.videos` to only work for _your own_ channel, whereas now you can use it on _any_ channel.

I still want to be able to run `channel.videos` to get the list of videos of a channel that I don't own, since that information is _public_.

Also, the `has_many :videos` association of a channel is cached with an instance variable, so it does not invoke YouTube again every time is called. Therefore, if you run

```
channel.videos(q: 'foo')
channel.videos(q: 'bar')
```

then the second call would still return the results of the first call.

Therefore, my suggestion is to create a new `Yt::Search` model that reflects what you want to do.

```
account = Yt::Account.new access_token: '...'
search = Yt::Search.new q: 'laser', type: :video, auth: account
```

which would let you search **all videos on YouTube** that match 'laser'.

Is that what you want?

claudiofullscreen, line 24: Or I might just have misinterpreted everything, and you still want to loop through the videos of a specific channel, but only acts on those that match some filter?

Is this the line that you are trying to replace?
https://github.com/Fullscreen/Accounts/blob/master/lib/extensions/youtube_it.rb#L30

stavro, line 24: Ok.. I'll do that.

Btw, the `if @parent.auth == @auth` line would allow `channel.videos` to work on other people's channels (just not searching)... But yes, it should be more generic.  Ok.

claudiofullscreen, line 24: @stavro So probably we should have this:

```
channel.videos # return all the videos in a channel. Can be your channel or someone else’s

account.videos # return all the videos owned by the account

account.videos(q: 'laser') # return all the videos owned by the account, filtered by laser
```

Would that work?

stavro, line 24: Yep! I will go ahead and add that syntax.

claudiofullscreen, line 24: Let me think about it a little longer…

claudiofullscreen, line 24: Maybe it would be better like

```
account.videos.where(q: 'laser')
```

Let me see how that would work

ivankosdy:
Hi,

youtube api allows account videos filtering by video IDs,
but it does not work in your library :/

would it be possible to add this feature as well?

Thanks! :)

PR #10, opened 2014-06-13
Update docs badge in README

rrrene:
Hi there,

this patch updates the URL of the docs badge in your README. Inch's badges will be served from it's own CI service http://inch-ci.org in the future and I will slowly retire the old [Inch Pages project](http://inch-pages.github.io/). 

With the new service, people can [add projects straight from the homepage](http://inch-ci.org/) and also [configure a webhook to rebuild](http://inch-ci.org/howto/webhook) their badges auto-magically :star2:

I am so happy that Inch and Inch Pages were so well received in the community and that I am now able to keep [my promise](http://trivelop.de/2014/02/24/visibility-of-documentation/) to deliver a real, [open source](https://github.com/inch-ci) web service for the badges.

P.S. I have not written an announcement or anything for this, yet. **But**: Everybody who reads this is invited to add their projects to the site. Please help me find the last bugs, so I can start writing the "Introducing Inch CI" blog post :wink:


claudiofullscreen:
@rrrene Wow, your service improved a lot! One thing that would make it even better would be to compatible with http://shields.io – in this way developers could choose the best style of badge that shows in their page. For instance, I might want to display the % of documented code, since it might be easier to read out than an horizontal colored bar. Do you have plans to support that? In short, you'd only have to add another format (in addition to .png and .svg): for instance a .json that only has the percentage of documented code. Thanks!

rrrene:
I will certainly look into that in the future!

PR #19, opened 2014-07-03
New account.upload_video (both local and remote)

claudiofullscreen:
It is now possible for accounts to upload videos using the resumable
upload protocol.

This commit does not implement the entire protocol, just the simple
case where the upload "goes through" in one step. Still to be implemented
are the cases in which the upload is actually resumed.


jponc:
@KieranP good to see you here man!

@KieranP @claudiofullscreen Any plans on implementing chunk download to this gem?

EDIT: I saw this https://github.com/Fullscreen/yt/pull/120 keen on deploying this soon?

KieranP:
@jponc You might want to use the 90seconds gem. It contains a bunch of bug fixes and optimisations. https://github.com/90seconds/yt

jponc:
@KieranP Does it provide chunk uploads? I tried to do a bit of benchmark -- I uploaded a 400mb file to youtube and it only ate around 100mb memory. So I assume it's already doing chunk uploads out of the box and not loading the entire file in memory? But I can't find it exactly from the source

PR #24, opened 2014-07-16
Make channel.videos access more than 500 videos

### LINKED ISSUE #23: Channel.videos limited to 500
When accessing the videos of a channel with more than 500 videos, Channel.videos seems to be limited to 500 results. For example, for the channel raocow (channel = Yt::Channel.new url: 'http://www.youtube.com/raocow') channel.videos.count == 500 but channel.video_count == 3500 or so.

Googling makes it seem like this might be an API limit? I believe that youtube_it gets all videos for a channel by setting a start index and iterating until all videos are obtained. Is it possible to do such a thing with this library? Otherwise, would it be possible for this to be supported?

I considered writing such a thing myself but I was unsure where it fit in with the architecture of the library.

Thanks for the great software,
Cat

claudiofullscreen:
According to http://stackoverflow.com/a/23256768, YouTube API
won't provide more than ~500 search results for any arbitrary query.

This commit allows to overcome this limitation by splitting channel.videos
into multiple paginated queries that ask for videos in chronological order
and use the `publishedBefore` parameter to collect all the videos.

Closes #23

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    yt (0.7.8)
+    yt (0.7.9)
       activesupport
 
 GEM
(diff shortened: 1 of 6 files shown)


chuchuva:
channel.videos returns less than 500 videos for the channel [UCwmZiChSryoWQCZMIQezgTg](https://www.youtube.com/channel/UCwmZiChSryoWQCZMIQezgTg)

PR #27, opened 2014-07-22
Add Claims#list and ClaimSearch#list

stavro:
Adds the functionality to list Claims, and search for Claims.

https://developers.google.com/youtube/partner/docs/v1/claims/list

https://developers.google.com/youtube/partner/docs/v1/claimSearch/list

The way we handle attribute assignment is slightly different between each model class.  Some assign to instance variables, some don't.  I just kept this in line with some other models, but I could see some room for improvement on the claim attributes.

Thoughts?

### DIFF
--- a/lib/yt/collections/claim_searches.rb
+++ b/lib/yt/collections/claim_searches.rb
@@ -1,0 +1,24 @@
+require 'yt/models/claim'
+
+module Yt
+  module Collections
+    class ClaimSearches < Base
+
+    private
+
+      def new_item(data)
+        Yt::Claim.new data: data
+      end
+
+      # @return [Hash] the parameters to submit to YouTube to list claims that match the search criteria.
+      # @see https://developers.google.com/youtube/partner/docs/v1/claimSearch/list
+      def list_params
+        super.tap do |params|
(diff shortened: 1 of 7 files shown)


lib/yt/collections/claim_searches.rb
```diff
@@ -0,0 +1,24 @@
...
+      def new_item(data)
+        Yt::Claim.new data: data
+      end
+
+      # @return [Hash] the parameters to submit to YouTube to list claims that match the search criteria.
+      # @see https://developers.google.com/youtube/partner/docs/v1/claimSearch/list
+      def list_params
+        super.tap do |params|
+          params[:path] = "/youtube/partner/v1/claimSearch"
+          params[:params] = {onBehalfOfContentOwner: @parent.owner_name}.merge(@extra_params || {})
```
stavro, line 18: The use of `@extra_params` to enable the `.where` clause could use some DRYing up.

claudiofullscreen:
Closed in favor of https://github.com/Fullscreen/yt/pull/29 – if there’s anything in this PR that’s missing there, let's just fix that one!

PR #29, opened 2014-07-22
Add content_owner.claims to list Content ID claims

claudiofullscreen:
Add content_owner.claims to list Content ID claims

A new Claim class is introduced to list YouTube ContentID Claims.
See https://developers.google.com/youtube/partner/docs/v1/claims

content_owner.claims also accepts `where` to only list claims that
match a set of IDs, such as:

```
content_owner.claims.where id: 'A1234,B1345'
content_owner.claims.where asset_id: 'A123'
content_owner.claims.where video_id: 'x1234'
content_owner.claims.where q: 'something in the title'
```


lib/yt/models/claim.rb
```diff
@@ -0,0 +1,122 @@
...
+        status == 'potential'
+      end
+
+      # @return [Boolean] whether the claim is takedown.
+      def takedown?
+        status == 'takedown'
+      end
+
+      # @return [Boolean] whether the claim status is unknown.
+      def has_unknown_status?
```
claudiofullscreen, line 75: @stavro  I feel like `claim.unknown?` does not read as well as `claim.has_unknown_status?`. But the other methods work just fine, for instance `claim.disputed?`

lib/yt/models/claim.rb
```diff
@@ -0,0 +1,122 @@
...
+      end
+
+      # @return [Boolean] whether the covers the audiovisual of the content.
+      def audiovisual?
+        content_type == 'audiovisual'
+      end
+
+      # @return [Time] the date and time that the claim was created.
+      def created_at
+        @created_at ||= Time.parse @data["timeCreated"]
```
claudiofullscreen, line 103: I suppose `timeCreated` is always included in the response, so I don't bother rescuing nil

lib/yt/models/claim.rb
```diff
@@ -0,0 +1,122 @@
...
+      end
+
+      # Return whether the video should be blocked where not explicitly owned.
```
claudiofullscreen, line 106: I like to copy the documentation from YouTube, in case they change it, and also to make it available on rubydoc.info

spec/requests/as_content_owner/content_owner_spec.rb
```diff
@@ -4,5 +4,7 @@
 describe Yt::ContentOwner, :partner do
   # NOTE: Uncomment once size does not runs through *all* the pages
   # it { expect($content_owner.partnered_channels.size).to be > 0 }
+  # it { expect($content_owner.claims.size).to be > 0 }
   it { expect($content_owner.partnered_channels.first).to be_a Yt::Channel }
+  it { expect($content_owner.claims.first).to be_a Yt::Claim }
```
claudiofullscreen: This integration test does not actually guarantee that YouTube returns all those fields for each claim, but I figure that kind of test is well suited for a Claim/get (when we write that)

lib/yt/collections/claims.rb
```diff
@@ -0,0 +1,28 @@
...
+      def new_item(data)
+        Yt::Claim.new data: data
+      end
+
+      # @return [Hash] the parameters to submit to YouTube to list claims
+      #   administered by the content owner.
+      # @see https://developers.google.com/youtube/partner/docs/v1/claims/list
+      def list_params
+        super.tap do |params|
+          params[:params] = {onBehalfOfContentOwner: @parent.owner_name}
```
stavro: Is there a reason you removed the `extra_params` from here?  This method allows some parameters (like specifying the `id`)

claudiofullscreen: I thought that was part of `search`. Is it used by `list` too?

stavro: It has limited use for `list`.. but yes it could be used to retrieve a particular batch of claims if you pass `.where(id: "1,2,3,4")`

lib/yt/models/claim.rb
stavro, line 75: :thumbsup:

spec/requests/as_content_owner/content_owner_spec.rb
```diff
@@ -2,7 +2,27 @@
...
+        let(:claim_id) { ENV['YT_TEST_PARTNER_CLAIM_ID'] }
+        it { expect(count).to be > 0 }
+      end
+
+      context 'given an unknown claim ID' do
+        let(:claim_id) { '--not-a-matching-claim-id--' }
+        it { expect(count).to be_zero }
+      end
+    end
+  end
```
claudiofullscreen, line 87: @stavro I've restored the option to call `content_owner.claims.where(id: '123')` and added a test.

It just feels weird that the parameter is called `id` but the doc says:

> The id parameter specifies a list of comma-separated YouTube claim IDs to retrieve.

It should be called `ids`! :sleepy:

stavro, line 87: Looks good to me :sheep:

claudiofullscreen:
@stavro This should have everything, documented and tested. You can now do any of the following:

```
content_owner.claims.where id: 'A1234,B1345'
content_owner.claims.where asset_id: 'A123'
content_owner.claims.where video_id: 'x1234'
content_owner.claims.where q: 'something in the title'
```

If you're good, I'll just wait for Travis and coverall to be happy, then merge

lib/yt/collections/claims.rb
```diff
@@ -0,0 +1,46 @@
...
+      def new_item(data)
+        Yt::Claim.new data: data
+      end
+
+      # @return [Hash] the parameters to submit to YouTube to list claims
+      #   administered by the content owner.
+      # @see https://developers.google.com/youtube/partner/docs/v1/claims/list
+      # @see https://developers.google.com/youtube/partner/docs/v1/claimSearch/list
+      def list_params
+        (@extra_params ||= {}).slice! :id, :asset_id, :video_id, :q
```
stavro: I don't really see a need to be explicit about the parameters, if the user wants to send a new parameter to the API, let them.  (examples: new parameters are introduced, a trace token is given to us, etc etc).

By slicing these four parameters out aren't you also preventing any of the optional parameters from being sent?

image

lib/yt/collections/claims.rb
```diff
@@ -0,0 +1,46 @@
...
+
+        super.tap do |params|
+          params[:params] = claims_params
+          params[:path] = claims_path
+        end
+      end
+
+      def claims_params
+        {onBehalfOfContentOwner: @parent.owner_name}.tap do |params|
+          (@extra_params ||= {}).each do |key, value|
```
claudiofullscreen, line 31: @stavro Look ma', no more `slice`!

spec/requests/as_content_owner/content_owner_spec.rb
```diff
@@ -2,7 +2,87 @@
...
+
+          describe 'applies the filter if the filter name is under_scored' do
+            let(:params) { {q: query, created_before: remote_date} }
+            it { expect(first).not_to be }
+          end
+
+          describe 'applies the filter if the filter name is camelCased' do
+            let(:params) { {q: query, createdBefore: remote_date} }
+            it { expect(first).not_to be }
+          end
```
claudiofullscreen, line 78: Some people like :camel: and some people don't, but we :heart: all of them!

PR #30, opened 2014-07-24
Add account.content_owners to list Content Owners

stavro:
Add account.content_owners to list Content Owners for an account

A new ContentOwner collection class is introduced to list Content Owners.
See https://developers.google.com/youtube/partner/docs/v1/contentOwners/list

``` ruby
yt_account.content_owners.count
  => 3
yt_account.content_owners.map &:id
  => ["cms_name_1", "cms_name_2", ...]
```

### DIFF
--- a/lib/yt/collections/content_owners.rb
+++ b/lib/yt/collections/content_owners.rb
@@ -1,0 +1,32 @@
+require 'yt/collections/base'
+require 'yt/models/content_owner'
+
+module Yt
+  module Collections
+    # Provides methods to interact with a collection of Content Owners.
+    #
+    # Resources with content_owners are: {Yt::Models::Account accounts}.
+    class ContentOwners < Base
+
+    private
+
+      def new_item(data)
+        Yt::ContentOwner.new owner_name: data['id'], authentication: @auth.authentication
+      end
+
+      # @return [Hash] the parameters to submit to YouTube to list content
+      #   owners administered by the account.
+      # @see https://developers.google.com/youtube/partner/docs/v1/contentOwners/list
+      def list_params
+        super.tap do |params|
+          params[:params] = content_owners_params
+          params[:path] = '/youtube/partner/v1/contentOwners'
+        end
+      end
+
+      def content_owners_params
+        {fetchMine: true}
+      end
+    end
+  end
+end
--- a/spec/requests/as_content_owner/account_spec.rb
+++ b/spec/requests/as_content_owner/account_spec.rb
@@ -1,0 +1,25 @@
+# encoding: UTF-8
+require 'spec_helper'
+require 'yt/models/account'
+
+describe Yt::Account, :partner do
+  subject(:account) { Yt::Account.new id: id, authentication: $content_owner.authentication }
+
+  describe '.content_owners' do
+    let(:content_owners) { account.content_owners }
+
+    context 'given a partenered account with content owners', :partner do
+      let(:id) { $content_owner.id }
+
+      specify 'returns the associated content owners' do
+        expect(content_owners.size).to be > 0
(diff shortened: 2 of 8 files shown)


lib/yt/models/content_owner.rb
```diff
@@ -15,8 +15,17 @@ class ContentOwner < Account
...
+      end
+
+      def display_name
+        @display_name ||= @data["displayName"]
+      end
+
       def initialize(options = {})
         super options
+        @data = options[:data] || {}
         @owner_name = options[:owner_name]
```
stavro, line 20: I have some concerns with the use of `owner_name`.  The use of the "friendly name" to query the youtube api using the `onBehalfOfContentOwner` is going away (see figure below), and after Q3 these parameters will only take an ID, and the "display name" is purely for aesthetic purposes.

Therefore, I propose (in a different PR), we do away with the idea of "owner name" as a unique identifier, and make it understood from the beginning it's a unique ID.

image

lib/yt/collections/content_owners.rb
```diff
@@ -0,0 +1,34 @@
...
+    # Provides methods to interact with a collection of Content Owners.
+    #
+    # Resources with content_owners are: {Yt::Models::Account accounts}.
+    class ContentOwners < Base
+
+    private
+
+      def new_item(data)
+        data[:owner_name] ||= data["id"]
+        Yt::ContentOwner.new data: data, access_token: @auth.access_token, refresh_token: @auth.refresh_token
```
stavro: There is no straightforward way (that I could see at least) to pass the authentication from one account to the next, without decomposing the parameters.

lib/yt/models/content_owner.rb
claudiofullscreen, line 20: When YouTube says "in Q3 2014", they actually mean "in Q3 2024" :stuck_out_tongue_closed_eyes:

I wouldn't worry at all about this now, especially given that they are not even serving the ID yet.

spec/requests/as_content_owner/account_spec.rb
```diff
@@ -0,0 +1,25 @@
+# encoding: UTF-8
+require 'spec_helper'
+require 'yt/models/account'
+
+describe Yt::Account, :partner do
+  subject(:account) { Yt::Account.new id: id, authentication: $content_owner.authentication }
```
claudiofullscreen, line 6: This looks better.

You know that `ContentOwner < Account`, therefore you could also end up writing:

```
content_owner.content_owners.first.content_owner.content_owners.first.content_owner
```

:stuck_out_tongue_winking_eye:

stavro, line 6: Yep! haha. I was actually doing that earlier.  I don't particularly think it's a bad thing though.

claudiofullscreen:
This is ready to go. The only difference compared to the description above is that you'll have to run

```
account.content_owners.map &:owner_name  #=> ["cms_name_1", "cms_name_2", ...]
```

and not  `account.content_owners.map &:id`.

When things change for YouTube API, we can keep the `id` for the 22-chars ID, but still keep the `owner_name` method available as it is now.

stavro:
I'm fine with the `owner_name`.. but I would like to be able to access the `displayName` that comes back from the API.. any thoughts?

claudiofullscreen:
You're right, ContentOwner has more fields now as described in
https://developers.google.com/youtube/partner/docs/v1/contentOwners#resource

I feel this PR is strong enough to be merged, so now you can just `gem 'yt' ~> 0.8.3` and start working with the new API endpoint.

Make a separate PR/issue to add the extra fields to ContentOwner, it won't take long, but for now… :sleeping: :wink:

stavro:
Thanks! :heart:

PR #31, opened 2014-07-24
[Discussion] #include method on base collection

stavro:
Throwing this out there for discussion...  I'm going to need to make _a lot_ of API calls to get the view_counts and such of videos, and I'd like a way to do that as efficiently as possible (both time and quota).

Each call to the API is one additional quota unit, as well as ~300ms waiting.  Through `yt` there is no way to do this in bulk yet.

This PR would allow me to get the statistics of batched videos.

What do you think about adding an `include` method to specify additional parts?

``` ruby
Yt::Collections::Videos.new.include(:statistics, :content_owner_details).where(id: 'K9XCKP9KN7A').first
```

### DIFF
--- a/lib/yt/collections/base.rb
+++ b/lib/yt/collections/base.rb
@@ -13,6 +13,8 @@ module Yt
       def initialize(options = {})
         @parent = options[:parent]
         @auth = options[:auth]
+        @extra_params = {}
+        @extra_parts = []
       end
 
       def self.of(parent)
@@ -22,6 +24,12 @@ module Yt
       def where(conditions = {})
         @items = []
         @extra_params = conditions
+        self
+      end
+
+      def include(*parts)
+        @items = []
+        @extra_parts = parts
         self
       end
     end
--- a/lib/yt/collections/claims.rb
+++ b/lib/yt/collections/claims.rb
@@ -28,7 +28,7 @@ module Yt
 
       def claims_params
         {onBehalfOfContentOwner: @parent.owner_name}.tap do |params|
-          (@extra_params ||= {}).each do |key, value|
+          @extra_params.each do |key, value|
             params[key.to_s.camelize :lower] = value
           end
         end
--- a/lib/yt/collections/videos.rb
+++ b/lib/yt/collections/videos.rb
@@ -15,7 +15,7 @@ module Yt
       #   the items returned by asking YouTube for a list of videos.
       # @see https://developers.google.com/youtube/v3/docs/videos#resource
       def new_item(data)
-        Yt::Video.new id: data['id']['videoId'], snippet: data['snippet'], auth: @auth
+        Yt::Video.new id: data['id']['videoId'], snippet: data['snippet'], statistics: data['statistics'], auth: @auth
       end
 
       # @return [Hash] the parameters to submit to YouTube to list videos.
@@ -23,7 +23,7 @@ module Yt
       def list_params
         super.tap do |params|
           params[:params] = @parent.videos_params.merge videos_params
-          params[:path] = '/youtube/v3/search'
+          params[:path] = videos_path
         end
       end
 
@@ -43,10 +43,21 @@ module Yt
       end
 
       def videos_params
-        params = {type: :video, maxResults: 50, part: 'snippet', order: 'date'}
+        params = {type: :video, maxResults: 50, part: videos_parts, order: 'date'}
         params[:publishedBefore] = @published_before if @published_before
-        @extra_params ||= {}
         params.merge @extra_params
+      end
+
+      def videos_path
+        if @extra_params.empty? || @extra_params.key?(:id)
+          '/youtube/v3/videos'
+        else
+          '/youtube/v3/search'
+        end
+      end
+
+      def videos_parts
+        [:snippet].concat(@extra_parts).uniq.join(',')
       end
     end
   end
--- a/lib/yt/models/resource.rb
+++ b/lib/yt/models/resource.rb
@@ -20,6 +20,7 @@ module Yt
         @auth = options[:auth]
         @snippet = Snippet.new(data: options[:snippet]) if options[:snippet]
         @status = Status.new(data: options[:status]) if options[:status]
+        @statistics_set = StatisticsSet.new(data: options[:statistics]) if options[:statistics]
       end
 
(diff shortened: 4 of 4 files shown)


claudiofullscreen:
:yellow_heart: I think this goes in the right direction! Should we rename Yt to _ActiveYoutube_? :grinning: 

There's a couple of gotchas in this PR that I want to go through before merging.

Also (and I always make the same mistake with ActiveRecord) the method should be called `includes`, not `include` to [mimick ActiveRecord’s syntax](http://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-includes).

claudiofullscreen:
@stavro and @Nuru – I finally found myself with a similar problem, and I'd like to know what you think of my idea.

In ActiveRecord, you can use `select` to specify which fields to return from the same table that you are querying, and you can use `includes` to load data from a different table, but making **one** request for the entire collection, rather than one per item, hence avoiding N+1 queries.

What I'm working on now is to obtain the list of channels partnered with a content owner, and gather data about them with the least amount of queries. In short:

``` ruby
yt_content_owner.partnered_channels.each do |yt_channel|
  channel = Channel.find_or_initialize_by slug: yt_channel.id
  channel.reach = yt_channel.subscriber_count
  channel.demographics = yt_channel.viewer_percentages
  channel.save
end
```

---

The first N+1 query from the code above comes from the fact that `.partnered_channels` only retrieves the `snippet` and `status` parts of each channel. When I call `yt_channel.subscriber_count`, I want some data that is in the `statistics` part, so I need to run an additional query for each channel.

This is the same problem you faced in #84 and it's where the `select` syntax would make sense, because we want to run the same query (give me the list of channels partnered with the content owner) but specifying a different set of "columns" being returned (`statistics`, not just `snippet` and `status`).

---

The second N+1 query from the code above comes from the fact that `.partnered_channels` hits the YouTube Data API. When I call `yt_channel.viewer_percentages`, I'm asking for data that is available in the YouTube Analytics API instead, so I need to run an additional query for each channel.

This case is different because I cannot simply run the same query "with different columns". I **must run a separate query** (against Analytics API), but still I want a way to run only **one** query for the entire set of channels, not one query per channel.

The solution comes from [the August 28, 2014 YouTube announcement](https://developers.google.com/youtube/analytics/revision_history): you can now pass a list of channels to the Analytics API queries. Therefore the optimal process is:
- retrieve the **list of channels** with one query to the YouTube Data API
- retrieve the **list of viewer percentages** with one query to the YouTube Analytics API, by passing the list of channel IDs collected from the previous query.

---

Summing up, this is the syntax that I imagine would implement all of this:

``` ruby
yt_content_owner.partnered_channels.select(:statistics).includes(:viewer_percentages).each do |yt_channel|
  channel = Channel.find_or_initialize_by slug: yt_channel.id
  channel.reach = yt_channel.subscriber_count
  channel.demographics = yt_channel.viewer_percentages
  channel.save
end
```

As in ActiveRecord, I would use `select` to retrieve additional data from the same "table" the query is running against, and I would use `includes` to run an additional query that would include as a parameter all the channels, then parse the result and attach it to each individual channel.

Thoughts?

claudiofullscreen:
cc @jcohenho (This is for the problem we faced yesterday)

Nuru:
I'm not going to comment on the syntax because I don't like Active Record's syntax. 

I will point out that there are limits to the APIs and memory management issues to consider. The data API returns the result set in pages and returns at most 50 channels per page (usually fewer). (And for some reason occasionally fails to return a page.) The analytics API only accepts up to 200 channels/videos per request. So there is a mismatch there, and we don't want to simply take a page of requests from the data API and turn it into an analytics request because then we are making 4 times more analytics requests than we need to be making, wasting time and quota. 

The other thing is that we are on our way to having 60,000 channels, so we can't just buffer it all in memory prior to returning the results. 

So any solution to this problem is going to be at least slightly complicated. I'm in favor of having a solution, I just want to make sure it performs efficiently and handles YouTube's API quirks.

claudiofullscreen:
@Nuru thanks for your comment. I have two considerations:

1) By adding the `.includes(:viewer_percentages)` I'll make sure that all the channel ID in the current page will be submitted to the Analytics API to get their viewer percentages. This will decrease the number of queries immensely.

2) You mentioned that a "YouTube Data" page has 50 items, but a "YouTube Analytics" page accepts 200 items. Therefore with 1) we would still do 4 times the number of queries we need. To solve this, I think I will be able to add a parameter so you can specify how many pages to read at once. Right now, when you use Yt against the Data API, you always get (as a page) _one request of 50 items_ at the time (that's one query). Instead, I can make this into a parameter so you can specify that you always want _four requests of 50 items_

Nuru:
The details of the API such as 50 items vs 200 items should be hidden from
the gem client by the gem. The gem user should not need to worry about how
things are paginated behind the scenes. There are many reasons for this,
but one particular one is encapsulation: when the APIs change again we
should only have to change the gem, not the gem's clients, to respond to
the API changes.

claudiofullscreen:
@Nuru what you describe can already be implemented with Yt:

1) You can manually pass 50 channel IDs to the Channels collection, and you get their 50 snippets
2) You can manually pass 200 channel IDs to the ViewerPercentages collection, and you get their 200 viewer percentages

What you are asking for (if I understand correctly) is a way to **loop through a list of channels and their viewer percentages** at the same time (in the same loop); something like:

``` ruby
content_owner.partnered_channels.each do |channel|
  p channel.title
  p channel.viewer_percentages
end
```

The first time this loop is run, Yt has not loaded any channel, so it asks YouTube Data API for as many channels as it can: a page of 50. In this way, for the next 49 iterations, `p channel.title` will not require another API call.

Still during the first iteration, Yt has not loaded any viewer percentages, so it asks YouTube Analytics API for as many as it can. Since at this point Yt has already 50 `channel_id`, it can send all of them in one request, and for the next 49 iterations, `p channel.viewer_percentages` will not require another API call.

What do you suggest Yt should do instead by default (hidden by the client)? Should it instead make 3 more requests to the YouTube Data to also get the `channel_id` from 51 to 100, from 101 to 150 and from 151 to 200, and then making the request to YouTube Analytics API with the 200 channel ids?

Do you think that would be a better default?

PR #42, opened 2014-07-30
Add channel.content_owner_details

stavro:
The easiest way to confirm that a channel is in a particular network is by checking the content_owner_details of that channel.  The content owner details object is **only accessible** if you are the content owner.

https://developers.google.com/youtube/v3/docs/channels#contentOwnerDetails

``` ruby
content_owner = Yt::ContentOwner.new(access_token: "...", owner_name: "FullScreen")

# channel in FullScreen
channel = Yt::Channel.new(id: 'UC-Z3Pey219Tyr9aW1mX7HQQ', auth: content_owner)
channel.content_owner => "FullScreen"
channel.time_linked => 2013-03-14 01:14:54 UTC

# channel not in FullScreen
channel = Yt::Channel.new(id: 'UCcMTZY1rFXO3Rj44D5VMyiw', auth: content_owner)
channel.content_owner => nil
channel.time_linked => nil
```

Thoughts?

### DIFF
--- a/lib/yt/models/content_owner_detail.rb
+++ b/lib/yt/models/content_owner_detail.rb
@@ -1,0 +1,47 @@
+require 'yt/models/base'
+
+module Yt
+  module Models
+    # Encapsulates channel data that is relevant for YouTube Partners linked
+    # with the channel.
+    # @see https://developers.google.com/youtube/v3/docs/channels#contentOwnerDetails
+    class ContentOwnerDetail < Base
+      def initialize(options = {})
+        @data = options[:data] || {}
+      end
+
+      # Returns the name of the content owner linked to the channel.
+      #
+      # This method requires {Resource#auth auth} to return an authenticated
+      # instance of {Yt::ContentOwner} that can administer the channel.
+      # @return [String] if the channel is partnered with a content owner,
+      #   the name of the content owner linked to the channel.
+      # @return [nil] if the channel is not partnered with a content owner.
+      # @return [nil] if {Resource#auth auth} is a content owner without
+      #   permissions to administer the channel.
+      # @raise [Yt::Errors::Forbidden] if {Resource#auth auth} does not
+      #   return an authenticated content owner.
+      def content_owner
+        @content_owner ||= @data['contentOwner']
+      end
+
+      # Returns the date and time of when the channel was linked to the content
+      # owner.
+      #
+      # This method requires {Resource#auth auth} to return an authenticated
+      # instance of {Yt::ContentOwner} that can administer the channel.
+      # @return [Time] if the channel is partnered with a content owner,
+      #   the date and time when the channel was linked with the content owner.
+      # @return [nil] if the channel is not partnered with a content owner.
+      # @return [nil] if {Resource#auth auth} is a content owner without
+      #   permissions to administer the channel.
+      # @raise [Yt::Errors::Forbidden] if {Resource#auth auth} does not
+      #   return an authenticated content owner.
+      def linked_at
+        @linked_at ||= if @data['timeLinked']
+          Time.parse @data['timeLinked']
+        end
+      end
+    end
+  end
+end
--- a/spec/requests/as_account/channel_spec.rb
+++ b/spec/requests/as_account/channel_spec.rb
@@ -105,6 +105,11 @@ describe Yt::Channel, :device_app do
       expect{channel.viewer_percentages}.not_to raise_error
       expect{channel.viewer_percentage}.not_to raise_error
     end
+
+    it 'cannot give information about its content owner' do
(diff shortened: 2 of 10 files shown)


claudiofullscreen:
Oh yes, so many attributes still missing. I swear I'll write some meta-programming to automatically add all the attributes, just wait for it :grimacing: 

Meanwhile I'll check what's failing on this build

claudiofullscreen:
@stavro For now I've just rebased off master and `push --force` your commit.
Unfortunately YouTube API is still having problems today, so I cannot fully test this PR :crying_cat_face: 
Let's hope they can fix their issues today!

lib/yt/models/content_owner_detail.rb
```diff
@@ -0,0 +1,24 @@
...
+      # @return [String] The ID of the content owner linked to the channel.
+      def content_owner
+        @content_owner ||= @data['contentOwner']
+      end
+
+      # @return [String] The date and time of when the channel was linked to the
+      #   content owner.
+      def time_linked
+        @time_linked ||= Time.parse(@data['timeLinked']) rescue nil
+      end
```
claudiofullscreen, line 44: Let's call this `linked_at`, shall we? :neckbeard:

spec/requests/as_content_owner/channel_spec.rb
```diff
@@ -263,13 +263,20 @@
...
     context 'not managed by the authenticated Content Owner' do
       let(:id) { 'UCBR8-60-B28hp2BmDPdntcQ' }
 
-      it { expect{channel.earnings}.to raise_error Yt::Errors::Forbidden }
-      it { expect{channel.views}.to raise_error Yt::Errors::Forbidden }
+      specify 'information about its content owner cannot be retrieved' do
+        expect{channel.content_owner}.to raise_error Yt::Errors::Forbidden
+        expect{channel.linked_at}.to raise_error Yt::Errors::Forbidden
+      end
     end
```
claudiofullscreen, line 285: This is quite straightforward: if you are logged in as a CMS owner, but you try to retrieve the `content_owner` of a channel not owned by you, then you get `Yt::Errors::Forbidden`.

spec/requests/as_account/channel_spec.rb
```diff
@@ -105,6 +105,11 @@
       expect{channel.viewer_percentages}.not_to raise_error
       expect{channel.viewer_percentage}.not_to raise_error
     end
+
+    it 'cannot give information about its content owner' do
+      expect{channel.content_owner}.to raise_error Yt::Errors::Forbidden
+      expect{channel.linked_at}.to raise_error Yt::Errors::Forbidden
+    end
```
claudiofullscreen, line 112: @stavro How do you feel about this?

In short, if you are logged-in as a normal account (not a CMS owner) and try to retrieve the `content_owner` of a channel, you cannot access it, so YT raises `Yt::Errors::Forbidden`. 

In my opinion, this is more meaningful to YouTube behavior, which lets you _ask_ for the `contentOwnerDetails` part but then does not give it back to you if you are not a content owner. For instance:

```
curl -X GET -H "authorization: Bearer ..." "https://www.googleapis.com/youtube/v3/channels?id=UC...&part=contentOwnerDetails"
```

returns

```
{
 "kind": "youtube#channelListResponse",
 "etag": "\"gM...\"",
 "pageInfo": {
  "totalResults": 1,
  "resultsPerPage": 1
 },
 "items": [
  {
   "kind": "youtube#channel",
   "etag": "\"gM...\"",
   "id": "UC..."
  }
 ]
}
```

Basically, you are only asking YouTube for the `contentOwnerDetails` part of the channels, and YouTube does not give you any part back. The way in which I interpret this is: that part is _forbidden_ to you, therefore instead of just letting `channel.content_owner` and `channel.linked_at` return `nil` in the case above (which is confusing), I let Yt specifically raise `Forbidden`.

lib/yt/models/request.rb
```diff
@@ -28,6 +28,7 @@ def initialize(options = {})
       end
 
       def run
+        p as_curl
```
stavro: I've been adding this during my testing also... haha. I don't think you meant to commit it though

spec/requests/as_account/channel_spec.rb
stavro, line 112: Raising a `Forbidden` exception sounds fair to me

lib/yt/models/request.rb
claudiofullscreen: Good catch! And yes, I have to add a logger option

claudiofullscreen:
@stavro This PR will be pending until tests pass again, that is, until YouTube restores proper access the content-owner endpoints

claudiofullscreen:
@stavro this is now in ~> 0.9.8

PR #47, opened 2014-07-30
add claim.third_party? method

stavro:
Missed a method:

``` ruby
claim.third_party? => false
```


stavro:
Test failure is related to `channel.subscribe!` ... not this PR

PR #49, opened 2014-08-01
Add reverse, reverse_each and to_a method for Collections

Grafikart:


claudiofullscreen:
@Grafikart I'm not 100% sure about this. 

Collections use Enumerators as an abstraction of the fact that YouTube can provide multiple pages of results for a list. For instance, if you ask for a list of videos, and there are 2,000 videos matching your query, YouTube will return them in pages of 50 (as specified by `max_results`). Lists can even have millions of possible results.

That's why asking for the `last` element of such a list is not feasible, as it would require the entire list to be scanned (remember, possibly millions of items, or thousands of requests) just to get to the end of it.
If you look at YouTube API response, it does not even tell you _which page_ you are on; it just gives a token to simply go to the _next page_.

A solution to the above problem exists: for _some_ endpoints, YouTube API gives you an _order_ parameter that you can pass in the request. For instance [Search#list](https://developers.google.com/youtube/v3/docs/search/list) accepts an order parameter.

For instance, if you want to get the videos of a channel sorted by rating, rather than by relevance or date, you can type:

```
channel.videos.where(order: :rating).first
```

Once again, YouTube API is limited and, for instance, does not let you "reverse" the order, so if you wanted to get the video with the worst rating, you would still have to go through all of them.

Because of this, I wouldn't want to add methods like `last` or `reverse` to collections, because they would give the impression that you could _always_ get the last item (or reverse the order) of a list, while sometimes that might just take too long.

Grafikart:
Seems fair, so I would only add to_a so ppl can decide what to do with the result (I can fix the pull request with this method only if you want)

claudiofullscreen:
Could you show an example of what you are trying to achieve, to see if I can help?

I'm afraid that adding `to_a` would cause the same issues I described above: if you attach it to a collection that might have thousands or millions of results, then you would have to end indefinitely for the result to that method.

As I said, the problem is that the result that YouTube gives is not exactly "an array"; it's a list that you can only traverse in one direction. If you could explain your problem a little better, I can see if we can find a solution that does not block the caller in certain occasions.

PR #50, opened 2014-08-01
Added LiveStream attributes to video

Grafikart:
I wanted to list live videos with scheduled date in the future. It was unfortunately impossible since these attributes are in "liveStreamingDetails". I copied what you did for contentDetail to respect your pattern.

### DIFF
--- a/lib/yt/collections/live_streaming_details.rb
+++ b/lib/yt/collections/live_streaming_details.rb
@@ -1,0 +1,31 @@
+require 'yt/collections/base'
+require 'yt/models/live_streaming_detail'
+
+module Yt
+  module Collections
+    class LiveStreamingDetails < Base
+
+    private
+
+      # @return [Yt::Models::ContentDetail] a new content detail initialized
+      #   with one of the items returned by asking YouTube for a list of them.
+      def new_item(data)
+        Yt::LiveStreamingDetail.new data: data['liveStreamingDetails']
+      end
+
+      # @return [Hash] the parameters to submit to YouTube to get the
+      #   content detail of a resource, for instance a video.
+      # @see https://developers.google.com/youtube/v3/docs/videos#resource
+      def list_params
+        super.tap do |params|
+          params[:params] = live_streaming_details_params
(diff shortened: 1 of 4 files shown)


claudiofullscreen:
Thanks for the PR. I'll take a look at it right away :grin:

Grafikart:
basically it adds this four methods
actual_start_time
actual_end_time
scheduled_start_time
scheduled_end_time
concurrent_viewers
(I used delegate)

Grafikart:
I got problems while parsing date I added Time.zone.parse to avoid problem with timezone (youtube gives date in UTC)

claudiofullscreen:
@Grafikart I like your commit! I have added specs and description in the README.

Look at the commit https://github.com/Fullscreen/yt/commit/8386f14d2103d4019c819c7a869e7931180f349e

Do you agree with the changes I made to your code? I mainly fixed the documentation and dealt with the problem you mentioned above, of having YouTube sometimes returning `nil` values

Grafikart:
I'm fine with it, just the problem with dates and timezone.

With Time.parse 
time_tag @live.scheduled_start_time, will display the date using UTC

With Time.zone.parse
time_tag @live.scheduled_start_time, will display the date using the my rails timezone

Maybe it's a good idea to use zone to parse date to avoid additional work to display date ^^

claudiofullscreen:
@Grafikart I understand your concern about Time zones. 

Timestamps are always returned in UTC by YouTube API, and Yt respects that to mimic API, and also because it's a standard to what is expected from API.

However, if you want to convert a UTC timestamp in "your zone" in Rails, that's really easy to do in your Rails app, just use `in_time_zone`:

``` ruby
[4] pry(main)> utc_time = Time.parse "2014-08-01T17:52:00.000Z"
=> 2014-08-01 17:52:00 UTC
[6] pry(main)> utc_time.in_time_zone
=> Fri, 01 Aug 2014 10:52:00 PDT -07:00
```

Grafikart:
Ok so you can merge ^^

claudiofullscreen:
Closed by #50

PR #51, opened 2014-08-01
Added LiveStreamDetail attributes to video

claudiofullscreen:
Note: I made this PR adding documentation and specs to Grafikart’s code from #50 so I can remember about it once YouTube API goes back to functioning normally and tests stop failing because of YouTube API.

I won’t merge into master (and bump version) while tests fail… so until YouTube is back to normal, but the code looks good.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    yt (0.9.3)
+    yt (0.9.4)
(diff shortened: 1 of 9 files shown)


claudiofullscreen:
@Grafikart If you go update your Gemfile to `gem 'yt', '~> 0.9.4` you should see your changes applied!

Thanks again.

PR #52, opened 2014-08-01
Allow angle brackets when editing videos/playlists

claudiofullscreen:
Angle brackets are not characters allowed by YouTube, but instead
of failing, Yt replaces them with Unicode characters that look very
similar and are accepted by YouTube.

### DIFF
--- a/lib/yt/models/resource.rb
+++ b/lib/yt/models/resource.rb
@@ -1,3 +1,5 @@
+# encoding: UTF-8
+
 require 'yt/models/base'
 require 'yt/models/url'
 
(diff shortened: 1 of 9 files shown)


lib/yt/models/resource.rb
```diff
@@ -75,6 +86,17 @@ def underscore_keys!(hash)
         hash.dup.each_key{|key| hash[underscore key] = hash.delete key}
       end
 
+      # @return [Hash] the original hash with angle brackets characters in its
+      #   values replaced with similar Unicode characters accepted by Youtube.
+      # @see https://support.google.com/youtube/answer/57404?hl=en
+      def sanitize_brackets!(source)
+        case source
+          when String then source.gsub('<', '‹').gsub('>', '›')
```
stavro, line 94: You so fancy

claudiofullscreen:
Replace by #55 (more complete)

PR #54, opened 2014-08-01
Fix interpretation of video contentDetails duration field.

Nuru:
The `duration` field does not quite work as documented. This fixes the parsing of the duration field to handle the actual data returned.

### DIFF
--- a/lib/yt/models/content_detail.rb
+++ b/lib/yt/models/content_detail.rb
@@ -13,7 +13,7 @@ module Yt
 
       # @return [Integer] the duration of the video (in seconds).
       def duration
-        @duration ||= to_seconds @data.fetch('duration', 0)
+        @duration ||= youtube_duration_to_seconds @data.fetch('duration', 0)
       end
 
       # @return [Boolean] whether the video is available in 3D.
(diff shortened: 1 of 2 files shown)


claudiofullscreen:
@Nuru – first of all, thanks for the contribution. I never happened to stumble across such long videos, and I will gladly accept your contribution to make `duration` work for them as well.

As you can see, the file that you edited has a matching spec-file at https://github.com/Fullscreen/yt/blob/master/spec/models/content_detail_spec.rb#L7.
This file has different easy tests where I wrote what I expect to receive from YouTube, and what I expect the method to return, such as:

``` ruby
context 'given a content_detail with duration in hours, minutes, seconds' do
  let(:data) { {"duration"=>"PT1H18M52S"} }
  it { expect(content_detail.duration).to eq 4732 }
end
```

In your comments you mentioned three YouTube videos that you found and for which the current `duration` formula does not work. I invite you to add three tests to that spec-file, written as the one above, where you write what YouTube gives for those files, and you expect in return.

After you've done that on the current master, run the tests and make sure they all fail (as you mentioned). If not all of them fail, just leave the failing ones.

Once you've done that, replace the existing content of the `to_seconds`function with the one you suggest; re-run the tests and make sure they pass. Finally, update your commit with both code and tests and it will be ready to be merged.

Additionally, I suggest you do not rename the method; `to_seconds` still work, only the content needs to be fixed.

Thanks again!

Nuru:
OK, I added the tests. I want to leave the function name as `youtube_duration_to_seconds` because `to_seconds` is too vague and this conversion function is specific to how YouTube fills in this field. YouTube is (a) using the wrong specification and (b) not following the specification, so I want the name to warn people that this is not a generic conversion function.

claudiofullscreen:
Closed by #56 – @Nuru update your Gemfile to `gem 'yt', '~> 0.9.7'` and you should be ready to go!

PR #56, opened 2014-08-02
Also parse weeks and days in video’s duration

claudiofullscreen:
YouTube API does not fully respect ISO 8601 duration standard when
returning the "duration" of a video. This commit adds support for
videos with "days" and "weeks" in their duration which, although not
part of ISO 8601, are sometimes returned by YouTube API.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    yt (0.9.6)
+    yt (0.9.7)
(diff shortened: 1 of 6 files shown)


claudiofullscreen:
This PR reflects #54 with the following changes:
- more coherent yard-formatted documentation for the method
- more compact code, reusing the exiting RegEx for time-parsing
- bumping version of the gem

PR #57, opened 2014-08-03
.size doesn’t load all the pages of collection

claudiofullscreen:
Calling `size` on a collection (for instance on account.videos to know the
number of videos owned by an account) will now return the number of items as
specified in the "totalResults" field of the first page of the YouTube
response, rather than loading _all_ the pages (possibly thousands) and counting
exactly how many items are returned.

The old behavior is kept the same for the `count` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,31 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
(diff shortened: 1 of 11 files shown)


stavro:
Love it!

claudiofullscreen:
@stavro `~> 0.10.0` for this feature. 

Read the [CHANGELOG](https://github.com/Fullscreen/yt/blob/master/CHANGELOG.md#0100---2014-08-11) for details on how to update

PR #60, opened 2014-08-07
Add references.insert to references collection

stavro:
Adds the ability to create reference files via an existing claim.

@claudiofullscreen Don't attempt to test this without talking to me first.  Creating reference files in the CMS should not be done without caution.

This adds some shared functionality from Actions::List which should probably be DRY-ed up.

Allows for creation of references **from an existing claim** :

``` ruby
content_owner.references.insert(claim_id: 'abcd', content_type: 'audiovisual') #=> #<Yt::Models::Reference
```

### DIFF
--- a/lib/yt/actions/insert.rb
+++ b/lib/yt/actions/insert.rb
@@ -7,6 +7,9 @@ module Yt
     private
 
       def do_insert(extra_insert_params = {})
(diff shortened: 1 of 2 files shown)


claudiofullscreen:
@stavro Everything here should be taken care of by https://github.com/Fullscreen/yt/pull/67 – can you check if I missed anything? I didn't fully understand if the "extra params" in your commit would still be needed or not.

stavro:
Your PR looks good! Covers everything I need.

PR #65, opened 2014-08-12
Add `MatchPolicy` to update the policy of an asset

claudiofullscreen:
For instance, to change the policy of asset 'ASSET123' to 'POLICY456' run:

``` ruby
match_policy = Yt::MatchPolicy.new asset_id: 'ASSET123', auth: content_owner
match_policy.update policy_id: 'POLICY456'
```


claudiofullscreen:
@stavro `'~> 0.10.2'` and you are good to go!

PR #67, opened 2014-08-12
Add methods to insert and delete references

claudiofullscreen:
Content owners can now create references pointing to existing claims with

```
content_owner.create_reference claim_id: 'CLAIM123', content_type: 'audiovisual'
```

References can also be "soft" deleted with:

```
reference = Yt::Reference.new id: 'REF123', auth: content_owner
reference.delete
```

This "soft" deletion sets the status to "inactive" (YouTube does not provide a
method for really deleting a Reference).


stavro:
:thumbsup:  This should do it!

claudiofullscreen:
@stavro `gem 'yt', '~> 0.10.3'` has it!

PR #69, opened 2014-08-15
Add support for PATCH requests

claudiofullscreen:
Currently only YouTube Partners API supports PATCH, while YouTube
Data API still only has support for PUT requests to update resources.

The Patch action is introduced and used in the only Partner resource
that can get updated by Yt so far: MatchPolicy.


stavro:
:heart:

PR #71, opened 2014-08-16
Add Yt::Search.

pwnall:
Thank you very much for putting this gem together! I'm glad to see a gem that uses the V3 API!

I looked through the docs and code, and to the best of my knowledge, there's no way of doing a keyword search. The plumbing for using `GET /search` is all there, but it seems to only be exposed in the context of retrieving the videos in a channel or account.

This PR adds a Search model that can be used to run a keyword query. It's very bare-bones, because I wanted to get your opinion before I sink more time into it. At the same time, it is covered by tests, and I'm using it successfully in an application.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -397,6 +397,20 @@ item.delete #=> true
 
 *The methods above require to be authenticated as the playlist’s owner (see below).*
 
+Yt::Search
+----------
+
+Use [Yt::Search](http://rubydoc.info/github/Fullscreen/yt/master/Yt/Models/Search) to
+
+* find videos matching a user's query
+
+```ruby
+search = Yt::Search.new 'railsconf 2014 keynote'
+search.videos.first #=> #<Yt::Models::Video @id=...>
+```
+
+*The methods above do not require authentication.*
+
 
 Yt::Annotation
 --------------
--- a/lib/yt.rb
+++ b/lib/yt.rb
@@ -6,6 +6,7 @@ require 'yt/models/content_owner'
 require 'yt/models/match_policy'
 require 'yt/models/playlist'
 require 'yt/models/playlist_item'
+require 'yt/models/search'
 require 'yt/models/video'
 
 # An object-oriented Ruby client for YouTube.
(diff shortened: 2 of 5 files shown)


claudiofullscreen:
Hello @pwnall and thanks for your feedback! I appreciate it.

You are right, there is currently no way to search for videos without specifying either a parent account or a parent channel, but all the code is pretty much there.

Rather than creating a new `Yt::Search` model, I would like to take the approach of https://github.com/Fullscreen/yt/pull/72 

The example in your code searches for the first video matching 'railsconf 2014 keynote', where the results are ordered by relevance with a moderate safe search

With https://github.com/Fullscreen/yt/pull/72 that search would be run with the following commands:

``` ruby
videos = Yt::Collections::Videos.new auth: account
video = videos.where(q: 'railsconf 2014 keynote', order: 'relevance', safe_search: 'moderate').first
```

What do you think of this solution? Would it work for you?

pwnall:
Thank you for the quick response, @claudiofullscreen!

What you mentioned above would definitely work for me, as long as there's a way to do it with an api key instead of an account.

I think it'd be worth adding a search snippet to the README. No example talks about reaching into `Yt::Collections`, and there's a bit of magic happening there.

Thank you, and I look forward to using what you come up with!

claudiofullscreen:
@pwnall You are right!

https://developers.google.com/youtube/v3/docs/search/list does not require to be authenticated as an account, so the snippet of code can simply work as:

``` ruby
videos = Yt::Collections::Videos.new
video = videos.where(q: 'railsconf 2014 keynote', order: 'relevance', safe_search: 'moderate').first
```

Regarding the documentation, that’s what’s missing from #72 – I'll add tests, docs and examples in the README and then merge into master.

pwnall:
Thanks so much! :heart:

claudiofullscreen:
https://github.com/Fullscreen/yt/pull/72 takes care of this (as explained above).

@pwnall update your Gemfile to `'yt', '~> 0.11.1'` and you should be ready to go!

PR #75, opened 2014-08-19
added asset, assets, ownership, and claims.insert to support changing mo...

timzero:
Added some classes and methods needed to control monetization settings for an uploaded video.  The idea is to do as follows in Accounts:

```
asset = content_owner.create_asset type: 'web'
ownership = Yt::Ownership.new asset_id: asset.id, auth: content_owner
ownership.update kind: "youtubePartner#rightsOwnership", general: [{ratio: 100.0, owner: content_owner.owner_name, type: "exclude", territories: []}]
video = VaultVideo.find_by_slug 'famOnt'
policy = {rules: [{action: "monetize"}]}
content_owner.claims.insert assetId: asset.id, videoId: video.youtube_id, policy: policy, contentType: "audiovisual"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
+## 0.11.2 - 2014-08-19
+
+* [FEATURE] Add Assets Collection
+* [FEATURE] Add Asset model
+* [FEATURE] Add Ownership model
+* [FEATURE] Add VideoAdvertisingOptions model
+
 ## 0.11.1 - 2014-08-17
 
 * [ENHANCEMENT] Add Video search even without a parent account or channel
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ To install on your system, run
 
 To use inside a bundled Ruby project, add this line to the Gemfile:
 
-    gem 'yt', '~> 0.11.1'
+    gem 'yt', '~> 0.11.2'
 
 Since the gem follows [Semantic Versioning](http://semver.org),
 indicating the full version in your Gemfile (~> *major*.*minor*.*patch*)
--- a/lib/yt.rb
+++ b/lib/yt.rb
@@ -7,6 +7,7 @@ require 'yt/models/match_policy'
 require 'yt/models/playlist'
 require 'yt/models/playlist_item'
 require 'yt/models/video'
+require 'yt/models/ownership'
 
 # An object-oriented Ruby client for YouTube.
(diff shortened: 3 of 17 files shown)


claudiofullscreen:
@timzero Summing up, the use of this feature would be like this:

Say a user has uploaded a video with id 'VIDEO_ID' and say the user has picked the policy to Monetize everywhere, with id 'POLICY_ID', then when the user clicks on "Monetize!", the following would happen:

``` ruby
# We create an empty 'asset' that we will need to claim the video
asset = content_owner.create_asset type: 'web'
# We retrieve the initially empty 'ownership' that YouTube automatically assigns to that asset
ownership = Yt::Ownership.new asset_id: asset.id, auth: content_owner
# We specify that the rights for that asset belongs 100% everywhere to us
ownership.update general: [{ratio: 100.0, owner: content_owner.owner_name, type: "exclude", territories: []}]
# We specify that the uploaded video uses that asset, and that every usage of that asset must be monetized
content_owner.claims.insert asset_id: asset.id, video_id: 'VIDEO_ID', policy_id: 'POLICY_ID', content_type: "audiovisual"
```

Does it make sense?

timzero:
Yes, that makes sense to me.

On Tue, Aug 19, 2014 at 2:12 PM, claudiofullscreen <notifications@github.com

claudiofullscreen:
"I am a content owner and I want to own 100% of an asset in every territory" seems like a common pattern, so I'm thinking to add a method that replaces

``` ruby
asset = content_owner.create_asset type: 'web'
ownership = Yt::Ownership.new asset_id: asset.id, auth: content_owner
ownership.update general: [{ratio: 100.0, owner: content_owner.owner_name, type: "exclude", territories: []}]
```

with

``` ruby
asset = content_owner.create_asset type: 'web'
asset.ownership.obtain!
```

claudiofullscreen:
@timzero To better understand the last commit, what's a use case for that? 
Would it be something like the following for the `update`?

``` ruby
video.advertising_options.update ad_formats: %w(long trueview_inslate) 
```

What about the `get`? Could you provide an example? Thanks!

timzero:
@claudiofullscreen Yep, I was thinking something like this:

```
ad_options = Yt::VideoAdvertisingOptions.new video_id: 'Y_k6oPpZsOo', auth: content_owner
params = {}
params["adFormats"] = ["standard_instream", "trueview_instream", "overlay", "long"]
ad_options.update params
```

And the GET is just to populate the data before updating.  Might not be necessary if we use PATCH instead of UPDATE.

claudiofullscreen:
@timzero Update your Gemfile to `gem 'yt', '~> 0.11.2'` and you should be good to go!

See the description of #76 on how to write your code after that.

timzero:
:+1:

PR #77, opened 2014-08-21
Add asset#update via patch

stavro:
Allows for updating metadata fields of as asset

``` ruby
asset = Yt::Asset.new(id: 'A435476563920238', auth: content_owner)
asset.update metadata_mine: { customId: '7cj0MhObCQk' } #=> true
```


claudiofullscreen:
@stavro gem 0.11.3 – all yours!

PR #79, opened 2014-08-22
Add AS::Notifications to YouTube requests

stavro:
Opening this PR for discussion.

ActiveSupport::Notifications, by convention, should be namespaced right-to-left, so I've named this `request.yt` (feel free to change this though).

Could be useful for:
- Quota / Request analysis
- Time spent making YouTube requests
- Audit Trails

### DIFF
--- a/lib/yt/models/request.rb
+++ b/lib/yt/models/request.rb
@@ -51,8 +51,10 @@ module Yt
     private
 
       def response
-        @response ||= Net::HTTP.start(*net_http_options) do |http|
-          http.request http_request
+        @response ||= ActiveSupport::Notifications.instrument 'request.yt' do |payload|
+          payload[:method] = @method
+          payload[:request_uri] = uri
+          payload[:response] = Net::HTTP.start(*net_http_options) { |http| http.request http_request }
         end
       rescue OpenSSL::SSL::SSLError, Errno::ETIMEDOUT, Errno::ENETUNREACH, Errno::ECONNRESET => e
         @response ||= e
--- a/lib/yt/models/request.rb
+++ b/lib/yt/models/request.rb
@@ -51,13 +51,19 @@ module Yt
     private
 
       def response
-        @response ||= ActiveSupport::Notifications.instrument 'request.yt' do |payload|
-          payload[:method] = @method
-          payload[:request_uri] = uri
-          payload[:response] = Net::HTTP.start(*net_http_options) { |http| http.request http_request }
-        end
+        @response ||= send_http_request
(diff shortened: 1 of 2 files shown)


stavro:
Added documentation

lib/yt/models/request.rb
```diff
@@ -34,6 +34,12 @@ def initialize(options = {})
       def run
         p as_curl if Yt.configuration.developing?
 
+        ActiveSupport::Notifications.instrument('request.yt', host: @host, method: @method, path: @path, body: @body, query: @query) do |payload|
```
claudiofullscreen: Instead of passing a set of methods, wouldn't we be better off passing `request: self`?

lib/yt/models/request.rb
```diff
@@ -34,6 +34,12 @@ def initialize(options = {})
...
+          response # trigger request
```
claudiofullscreen: The fact that you had to add this extra line makes me think that what we really want to instrument is the whole `run` method, and not just this single line.

Maybe we can leave the `run` method as it is, and have the instrumentation _around_ the whole `run` method.

Does it make sense?

stavro: That does make sense, but I wanted to instrument each request, not lump together `401 unauthorized` + `refresh token` + `rerun`

claudiofullscreen: I see… then we could instrument the whole `response` method. Minus the memoization.

Or… we could instrument `Net::HTTP.request` directly, so you would get a log of _every_ HTTP request made by Yt.

lib/yt/models/request.rb
stavro: Let me look into this more - I'm pretty sure you want to treat the payload the same way you'd treat payload to a background job: should be simple values, not complex objects.

ActiveSupport::Notifications is built to use custom queuing libraries and I want to make sure this won't conflict.

lib/yt/models/request.rb
claudiofullscreen: Look at what [ActiveResource does](https://github.com/rails/activeresource/blob/82eb29ab023b3105b29bce31c9a00a3b9a9653aa/lib/active_resource/connection.rb#L118L122).

stavro: Yeah, either of those solutions would work for me.  Let me play with those and see what I come up with.

stavro: Ahh that's great. I like that!  Passing in the response could actually be very useful as well.

stavro:
Scratch that last comment, misread the code.  Just tried in this ActiveSupport 3 and it works. :thumbsup:

PR #80, opened 2014-08-23
Extract AS::Notification in a separate method

claudiob:
I like short methods and short lines of code.
@stavro how about extracting `instrument` into its own `send_http_request` method?

The rest looks great, I just need to run the test suite.
I think with this addition we might even get rid of that `p as_curl`!

### DIFF
--- a/lib/yt/models/request.rb
+++ b/lib/yt/models/request.rb
@@ -51,13 +51,19 @@ module Yt
     private
 
       def response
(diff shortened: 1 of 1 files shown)


stavro:
:thumbsup: :smile:

stavro:
`p as_curl` could still be useful for now, since this instrumentation doesn't provide easy access to the headers / auth tokens.  I wanted to keep this PR isolated from logging changes

PR #82, opened 2014-08-27
Videos#list id fix

stavro:
- [x] Only Videos#search should expected the nested `videoId` parameter

One minor change:

Tags are **only** visible under the following conditions:
1. You are using Videos#list, not Videos#search
2. You are using an authorized request, not a public one

If both of these conditions are met, you will receive the tags **if** you are authorized to see them (either your channel, or that channel's content owner).

### DIFF
--- a/lib/yt/collections/videos.rb
+++ b/lib/yt/collections/videos.rb
@@ -12,8 +12,14 @@ module Yt
     private
 
       def attributes_for_new_item(data)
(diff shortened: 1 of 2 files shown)


claudiofullscreen:
Replaced by #83

claudiofullscreen:
@stavro `0.11.5` !

PR #83, opened 2014-08-27
Fix video list

claudiofullscreen:
Closes #82 – The logic of tags in #82 was not right; we used to specify `includes_tags: false` when we were **sure** the response did not contain tags, but with the new endpoint this has changed.

Simply, when we ask for `video.tags` and the snippet _already has tags_ (because they were already returned by Videos#list), we return them, rather than erasing the `@snippet` variable.

On the other hand, when the snippet _does not have tags_ but we don’t have an authentication, it's useless to make an extra Video#show call, because without an authorizations, tags won't be returned anyways.


claudiofullscreen:
@stavro `0.11.5` !

PR #84, opened 2014-08-29
Add support for Collections::Videos.includes

stavro:
Allows for requesting multiple resource parts in a single video request.
### Usage

``` ruby
videos = Yt::Collections::Videos.new.where(id: 'sMeEwjQFGOU,BopDHDQGud8').includes(:statistics, :content_details, :status).each do |video|
  puts video.title
  puts video.view_count
  puts video.duration
  puts video.has_public_stats_viewable?
end
```
### Output

```
Successful YT get "https://www.googleapis.com/youtube/v3/videos?id=sMeEwjQFGOU%2CBopDHDQGud8&maxResults=50&order=date&part=statistics%2CcontentDetails%2Cstatus%2Csnippet&type=video&key=" in 241.749ms
MAN OF STEEL - Official Trailer #3 (With Henry Cavill Intro) (2013) [HD]
262984
196
true
Human Water Catapult - 55 Foot Launch! In 4k
750178
182
true
```

### DIFF
--- a/lib/yt/actions/base.rb
+++ b/lib/yt/actions/base.rb
@@ -10,6 +10,7 @@ module Yt
       # invoke transform_keys!{|key| key.to_s.underscore.to_sym}
       def underscore_keys!(hash)
         hash.dup.each_key{|key| hash[underscore key] = hash.delete key}
+        hash
       end
 
       def underscore(value)
--- a/lib/yt/collections/base.rb
+++ b/lib/yt/collections/base.rb
@@ -13,6 +13,8 @@ module Yt
       def initialize(options = {})
         @parent = options[:parent]
         @auth = options[:auth]
+        @where_params = {}
+        @included_parts = []
       end
 
       def self.of(parent)
@@ -41,10 +43,17 @@ module Yt
         end
       end
 
+      def includes(*parts)
+        self.tap do
+          @items = []
+          @included_parts = parts
+        end
+      end
+
     private
 
       def apply_where_params!(params = {})
-        params.merge!(@where_params ||= {})
+        params.merge! @where_params
       end
     end
   end
--- a/lib/yt/models/resource.rb
+++ b/lib/yt/models/resource.rb
@@ -22,6 +22,8 @@ module Yt
         @auth = options[:auth]
         @snippet = Snippet.new(data: options[:snippet]) if options[:snippet]
         @status = Status.new(data: options[:status]) if options[:status]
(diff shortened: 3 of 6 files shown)


lib/yt/actions/base.rb
```diff
@@ -10,6 +10,7 @@ module Base
       # invoke transform_keys!{|key| key.to_s.underscore.to_sym}
       def underscore_keys!(hash)
```
stavro, line 11: Not using this in this PR, but this method behaves unexpectedly by returning the **duped original** (not underscored) hash as the return value, rather than the new underscored value.  (I know it modifies the original, but the return value is still unexpected).

This change will bring it inline with `Hash#symbolize_keys!`

lib/yt/collections/base.rb
```diff
@@ -13,6 +13,8 @@ class Base
       def initialize(options = {})
         @parent = options[:parent]
         @auth = options[:auth]
+        @where_params = {}
```
stavro, line 16: I hate nil-checking throughout the code, can we just set this in the initializer for all collections?

lib/yt/actions/base.rb
claudiofullscreen, line 11: That method was not meant to have a meaningful return value.

We can change that, but when we do that, I suggest we also:
- change [camelize_keys!](https://github.com/Fullscreen/yt/blob/master/lib/yt/models/request.rb#L147) too, so the two methods behave in the same way
- benefit from this change and replace uses of these methods such as:

``` ruby
def update(attributes = {})
  underscore_keys! attributes
  do_patch body: attributes
  true
end
```

with

``` ruby
def update(attributes = {})
  do_patch body: underscore_keys!(attributes)
  true
end
```

lib/yt/models/resource.rb
```diff
@@ -22,6 +22,8 @@ def initialize(options = {})
         @auth = options[:auth]
         @snippet = Snippet.new(data: options[:snippet]) if options[:snippet]
         @status = Status.new(data: options[:status]) if options[:status]
+        @statistics_set = StatisticsSet.new(data: options[:statistics]) if options[:statistics]
+        @content_detail = ContentDetail.new(data: options[:content_details]) if options[:content_details]
```
stavro, line 26: Using `.where(part:'snippet,statistics')` would work if we add these two lines in, and the lines at the top of `collections/videos.rb`

stavro, line 26: If you'd prefer that syntax then that's fine with me.. just let me know

PR #85, opened 2014-09-10
Pass statistics and content details to Video#new

claudiofullscreen:
If a user specifies .where(part: 'statistics') when listing videos,
then YouTube will return the "statistics" part and the Video object
will be initialized with that data. Same for content details.

This **overrides the default behavior** of including status and snippet,
so if those parts are still required, the request should be:

```
.where(part: 'statistics,snippet,status,contentDetails')
```

---

This partially solves the issue tackled by #84 adding missing specs and moving code from "Resources" to "Videos", since not all resources have statistics and content details (e.g., playlists and playlist items don't).

I think the general issue raised by #84 is still relevant (to add an `.include` method, rather than passing the parts to `.where`), but for the sake of unblocking that issue, I suggest for this PR to be merged.

Note that extending the behavior of this PR from Videos to Channels can be done by… duplicating the code :scream: – not a big deal because these methods are being rewritten in a different branch :blush:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
+## 0.12.2 - 2014-09-09
+
+* [ENHANCEMENT] Accept `part` in the `where` clause of Videos, so statistics and content details can be eagerly loaded.
+
 ## 0.12.1 - 2014-09-04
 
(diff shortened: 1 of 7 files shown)


claudiofullscreen:
```
Finished in 5 minutes 59 seconds (files took 0.50654 seconds to load)
693 examples, 0 failures

Coverage report generated for RSpec to /Users/claudiob/code/yt/coverage. 4914 / 4923 LOC (99.82%) covered.
```

Closes #84 if accepted. @stavro if you think this can do it as a first step, feel free to merge and `rake release`.

PR #92, opened 2014-09-19
Allow for insertion of claims with custom policy rules

stavro:
This should be a valid operation with `yt`: 

``` ruby
content_owner.claims.insert(
  asset_id: 'A1234',
  video_id: 'abcd',
  content_type: 'audiovisual',
  policy: {
    rules: [
      {
        action: 'block'
      }
    ]
  }
)
```

However, the `yt` gem currently only looks for a `policy_id`.

This PR allows for the submission of custom policy rules.


claudiofullscreen:
@stavro I mean, `~> 0.13.4` and you are good to go!

PR #94, opened 2014-10-02
Add support for VideoAdvertisingOptions#get

stavro:
Example usage:

_Adding a third party ad server id to an existing YouTube video_:

``` ruby
  content_owner = Yt::ContentOwner.new refresh_token: '...'
  video_id = 'video_id'
  third_party_video_id = 'third_party_video_id'

  yt_video = Yt::Video.new(id: video_id, auth: content_owner)
  new_ad_formats = Set.new(yt_video.ad_formats).add('third_party').to_a
  yt_video.advertising_options_set.update(ad_formats: new_ad_formats, tp_ad_server_video_id: third_party_video_id)
```


claudiofullscreen:
@stavro With the existing code you can already
1. Update the ad formats of a video:
   
   ``` ruby
   ad_options = Yt::AdvertisingOptionsSet.new video_id: 'video_id', auth: content_owner
   ad_options.update ad_formats: %w(standard_instream long)
   ```
2. Update the third-party ad-server video ID
   
   ``` ruby
   ad_options = Yt::AdvertisingOptionsSet.new video_id: 'video_id', auth: content_owner
   ad_options.update tp_ad_server_video_id: 'third_party_video_id'
   ```

The only thing that you cannot do is to know what are the existing `ad_formats` for a certain video.

If the only goal of knowing that is to be able to add new ad formats and make a unique array, then I suggest that we just add a method to `Video` called `add_ad_formats`:

``` ruby
video = Yt::Video.new id: 'video_id', auth: content_owner
video.add_ad_formats 'long', 'third_party'
```

which would be defined in `video.rb` somehow like:

``` ruby
def add_ad_formats(*new_ad_formats)
  ad_formats = (advertising_options_set.ad_formats + new_ad_formats).uniq
  advertising_options_set.update ad_formats: ad_formats
end
```

Would that work?

stavro:
Adding a convenience method for `add_ad_formats` seems unnecessary in my opinion..   Where do you draw the line between having convenience methods for every action vs accessing the attributes themselves and letting the consumer of `yt` decide what to do with them?

claudiofullscreen:
@stavro Ok. Then if we are leaving access to `video.advertising_options_set` (to run commands like `update`) we can remove the following delegation, sounds ok?

``` ruby
delegate :ad_formats, :break_positions, :ad_breaks, :tp_ad_server_video_id, 
  :tp_targeting_url, :tp_url_parameters, :auto_generated_breaks, to: :advertising_options_set
```

stavro:
Removing the delegations is fine :+1:

PR #97, opened 2014-10-08
Allow statistics part on channels + viewer percentages

claudiofullscreen:
@stavro @Nuru This stems from my comment https://github.com/Fullscreen/yt/pull/31#issuecomment-58373026 so read that first.

In order to achieve what I describe there (having `select` and `includes` working on `.partnered_channels`), we first need the related Collection to accept those parameters and eager-load the requested data. This is what the current PR provides.

After the current PR, you will be able to do the following:

``` ruby
yt_channels = yt_content_owner.partnered_channels.where(part: 'statistics')

ids = "contentOwner==#{yt_content_owner.owner_name}"
filters = "channel==#{yt_channels.map(&:id).join ','}"
viewer_percentages = Yt::Collections::ViewerPercentages.new(auth: content_owner).where ids: ids, filters: filters

yt_channels.each do |yt_channel|
  channel = Channel.find_or_initialize_by slug: yt_channel.id
  channel.reach = yt_channel.subscriber_count
  channel.demographics = viewer_percentages[yt_channel.id]
  channel.save
end
```

If the number of partnered channels is <= 50, then this works perfectly, and **only uses two queries**.
However, as @Nuru pointed out in https://github.com/Fullscreen/yt/pull/31#issuecomment-58394525 this approach needs to be slightly adapted if there are more partnered channels, because:
- we cannot collect all their IDs at once with `yt_channels.map(&:id)` (they are paginated)
- we cannot ask YouTube Analytics API for more than 200 channels in each request

Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,9 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
-## 0.13.6 - unreleased
+## 0.13.6 - 2014-10-08
 
+* [ENHANCEMENT] Accept `includes(:viewer_percentages)` in `.partnered_channels` to eager-load multiple viewer percentages at once.
 * [ENHANCEMENT] Accept `where` in ViewerPercentages to collect data for multiple channels at once.
 * [ENHANCEMENT] Accept `part` in the `where` clause of Channels, so statistics can be loaded at once.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
+## 0.13.6 - unreleased
+
+* [ENHANCEMENT] Accept `part` in the `where` clause of Channels, so statistics can be loaded at once.
+
 ## 0.13.5 - 2014-10-06
 
 * [ENHANCEMENT] Add `advertising_options_set` and `ad_formats` to video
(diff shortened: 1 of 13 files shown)


claudiofullscreen:
I finally added `.includes(:viewer_percentages)`, so the following code can now be written and finally it only **runs 2 queries for every 50 channels**!!

``` ruby
content_owner.partnered_channels.where(part: 'statistics').includes(:viewer_percentages).each do |yt_channel|
  channel = Channel.find_or_initialize_by slug: yt_channel.id
  channel.reach = yt_channel.subscriber_count
  channel.demographics = viewer_percentages[yt_channel.id]
  channel.save
end

```

stavro:
Love it!

Since we're being competitive with our optimizations.. here you go :wink:

**1.25 requests per 50 channels** ;)

``` ruby
yt_channels = yt_content_owner.partnered_channels.where(part: 'statistics')

ids = "contentOwner==#{yt_content_owner.owner_name}"

yt_channels.each_slice(200) do |yt_channel_grouping|
  filters = "channel==#{yt_channel_grouping.map(&:id).join ','}"
  viewer_percentages = Yt::Collections::ViewerPercentages.new(auth: content_owner).where ids: ids, filters: filters

  yt_channel_grouping.each do |yt_channel|
    channel = Channel.find_or_initialize_by slug: yt_channel.id
    channel.reach = yt_channel.subscriber_count
    channel.demographics = viewer_percentages[yt_channel.id]
    channel.save
  end
end
```

claudiofullscreen:
@stavro That's exactly what @Nuru needs! So I would say:
- the current PR goes into master, and by default you run 2 queries for every 50 channels
- when I have time, I'll add extra pagination parameters, so you can set `maxResults` (now it's hard-coded to 50) and the number of requests to run per page
- meanwhile, @Nuru can use the code above

Agree?

stavro:
:+1:

Nuru:
I need this on video analytics even more than I do on channel analytics.

PR #105, opened 2014-12-11
Update collections/videos.rb default order

KevinVerre:
Previously, We were to ordering the search results for a query to be 'date' by default. (as in, order by most recently added videos). Change it to be order: 'relevance' instead. That is the what YouTube sorts by by default, and that is what most people expect to get when they do a search using the Yt gem. If they want newest they should explicitly ask for it.

### DIFF
--- a/lib/yt/collections/videos.rb
+++ b/lib/yt/collections/videos.rb
@@ -55,7 +55,7 @@ module Yt
           params[:type] = :video
           params[:max_results] = 50
           params[:part] = 'snippet'
-          params[:order] = 'date'
+          params[:order] = 'relevance'
           params[:published_before] = @published_before if @published_before
           params.merge! @parent.videos_params if @parent
           apply_where_params! params


claudiofullscreen:
The reason why `date` is by default is that it's the only way to go beyond YouTube API's pagination limit.

I think you can still change this default to `relevance` in your code by passing a parameter, right?

KevinVerre:
Yes, passing in relevance as a parameter works. However, I was surprised and briefly confused that it wasn't the default.

What do you mean by pagination limit? How were you encountering it? I'm still pretty new to using the YouTube API.

Your documentation doesn't explicitly say how to limit the number of results for a query. If I only care about the first 25 results, of course I don't want YouTube to bandwidth by sending me 500. I believe how it works is you send one request and YouTube is sending back one response for each page of results. At first I thought maxResults was a parameter for total number of results but then I realized it was for maxResults per page. I think there should be a parameter in Yt::Collections.videos.where that limits the total number of results. I did notice that if I call first(25) on my Collection object, it appears to only send one request (instead of twenty to get all 500 results). So the Yt gem's collection appears to be smart about how it sends requests and returns results. But you don't mention that in the ReadMe.

claudiofullscreen:
I'll make the documentation clearer.

PR #128, opened 2015-03-24
Only require core_ext/hash from ActiveSupport

jrostand:
This changes only two other files where `String#in?` was used. They now use `Array#include?`.

### DIFF
--- a/lib/yt/collections/reports.rb
+++ b/lib/yt/collections/reports.rb
@@ -57,7 +57,7 @@ module Yt
       end
 
       def rescue?(error)
-        'badRequest'.in?(error.reasons) && error.to_s =~ /did not conform/
+        error.reasons.include?('badRequest') && error.to_s =~ /did not conform/
       end
     end
   end
--- a/lib/yt/models/configuration.rb
+++ b/lib/yt/models/configuration.rb
@@ -57,13 +57,13 @@ module Yt
       # @return [Boolean] whether the logging output is extra-verbose.
       #   Useful when developing (e.g., to print the curl of every request).
       def developing?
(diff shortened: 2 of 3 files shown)


aharpole:
How I've missed your pull requests!

claudiofullscreen:
From ActiveSupport documentation:

1.1 Stand-Alone Active Support
In order to have a near-zero default footprint, Active Support does not load anything by default. It is broken in small pieces so that you can load just what you need, and also has some convenience entry points to load related extensions in one shot, even everything.

Thus, after a simple require like:

```
require 'active_support'
```

objects do not even respond to `blank?`.

claudiofullscreen:
My comment above is to say that the following line should not be removed:

```
require 'active_support'
```

On the other hand, what's the reason behind wanting to remove `require 'active_support/core_ext'` ?

jrostand:
@claudiofullscreen the issue I ran into was that including the whole `core_ext` module was making trouble in my app, which doesn't itself use ActiveSupport. It seemed logical to me to remove the more general `require` in the same stroke. Limited core extensions means less chance to interact negatively with an existing codebase.

claudiofullscreen:
Exactly what kind of trouble?

jrostand:
Among other things I was having tests fail for expecting certain objects' `#to_s` to return one value, but having AS change that. Sorry, but I can't give much more detail than that.

claudiofullscreen:
I'm sorry, but I need more details to consider this kind of pull request.

I don't want to preclude the chance to use other methods provided by ActiveSupport's core extensions without a valid and reproducible reason.

Feel free to re-open if you can provide more details.

PR #129, opened 2015-03-25
New `by: :traffic_source` option for reports

claudiofullscreen:
Before this commit, reports were always returning day-by-day data
(e.g. views were returned as `{Thu => 10.0, Wed => 11.0, ...}`).

After this commit, two channel reports (`views`, `estimated_minutes_watched`)
and one video report (`views`) can also return data by traffic source.

For instance, you can now run:

``` ruby
channel.views since: 7.days.ago, by: :traffic_source
```

and it would return something like:

``` ruby
{advertising: 10.0, related_video: 20.0, promoted: 5.0, subscriber: 1.0, channel: 3.0, other: 7.0}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ For more information about changelogs, check
 
 ## 0.14.0 - 2015-03-25
 
-* [FEATURE] New `:by` option for reports, to return results (e.g. `views`) either by day (default) or by traffic source.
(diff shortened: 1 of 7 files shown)


claudiofullscreen:
@brettu @stavro This is the first step for better reports!

brettu:
Awesome. :+1:

PR #136, opened 2015-04-08
Add claim_history collection and model

stavro:
- [x] Adds collection to obtain `claim_history` of claims
- [x] Add model for Yt::ClaimHistory
- [x] Add submodel for Yt::ClaimEvent

Example usage:

``` ruby
content_owner = Yt::ContentOwner.new(owner_name: '...')
claim = Yt::Claim.new(id: "1_8azDD5dXU", auth: content_owner)

history = claim.claim_history #=> #<Yt::Models::ClaimHistory
history.events[0].type #=> "dispute_create"
history.events[0].dispute_reason #=> "public_domain"
history.events[0].dispute_notes #=> "Acredito que seja um engano pois eu criei esse video utilizei minha voz no clipe, e a imagem do orkut que e publica."
```


claudiofullscreen:
Would it be possible to have an integration test (in `spec/requests`)?

That's the only way to prove that it actually does what YouTube says it does… since YouTube documentation about their API is never 100% reliable.

claudiofullscreen:
Or at least an addition to README or CHANGELOG with an example on how to use it.

stavro:
README and request spec added!

`ENV['YT_TEST_CLAIM_ID'] = 'i40f_S0e_Uo'` for testing

claudiofullscreen:
:+1:

PR #165, opened 2015-05-12
fix response parsing

maksymst:
with `json -> 1.8.2` any request is failing with error `ArgumentError: options must be a hash.
from /home/maxim/.rvm/gems/ruby-2.2.0@mytv/gems/json-1.8.2/lib/json/common.rb:468:in`parse'`

### DIFF
--- a/lib/yt/request.rb
+++ b/lib/yt/request.rb
@@ -183,7 +183,7 @@ module Yt
     def parse_response!
       response.body = case @response_format
         when :xml then Hash.from_xml response.body
-        when :json then JSON response.body
+        when :json then JSON response.body, {}
       end if response.body
     end
 


claudiofullscreen:
Hello @MaximSt and thanks for stopping by!

Could you copy/paste here the full command you are trying to run (except for personal API keys) and the full backtrace you are getting back? Thanks!

maksymst:
Hi @claudiofullscreen 

My configuration

``` ruby
Yt.configure do |config|
  config.client_id = ENV['GOOGLE_CLIENT_ID']
  config.client_secret = ENV['GOOGLE_CLIENT_SECRET']
  config.api_key = ENV['GOOGLE_API_KEY']
  config.log_level = :debug
end
```

I just took an example from description:

``` ruby
»  channel = Yt::Channel.new id: 'UCxO1tY8h1AhOz0T4ENwmpow'
=> #<Yt::Models::Channel:0x00000011f4dec0 @id="UCxO1tY8h1AhOz0T4ENwmpow", @auth=nil>
»  channel.title 
ArgumentError: options must be a hash.
from /home/maxim/.rvm/gems/ruby-2.2.0@mytv/gems/json-1.8.2/lib/json/common.rb:468:in `parse'
```

Problem happens in this method

``` ruby
  def JSON(object, *args)
    if object.respond_to? :to_str
      JSON.parse(object.to_str, args.first)
    else
      JSON.generate(object, args.first)
    end
  end
```

`JSON.parse(object.to_str, args.first)` requires second argument, but is not passed inside `parse_response!` method

claudiofullscreen:
Hello @MaximSt and thanks for your reply.

Given your feedback, I think the error is happening somewhere else, although I'm not sure where.
Here's why:

1) I tried running the code with the same configuration above (Ruby 2.2.0, Json 1.8.2) and it worked correctly

2) The error you reported says `ArgumentError: options must be a hash`. However, there is no variable called `options` in the `JSON` method

3) The fact that `parse_response!` does not pass a second argument means that the splatted `*args` is simply an empty array, therefore `args.first` is simply `nil`.

At this point, I'm curious to know more about this error you are experiencing. Have you tried `gem pristine json`? This will restore your JSON gem to its original form… maybe it was edited?

If you can't figure it out, send me more details, or a longer backtrace if you can. Thanks!

maksymst:
Hello @claudiofullscreen 

This is the longest backtrace I can get

```
ArgumentError: options must be a hash.
/home/maxim/.rvm/gems/ruby-2.2.0@mytv/gems/json-1.8.2/lib/json/common.rb:468:in `parse'
/home/maxim/.rvm/gems/ruby-2.2.0@mytv/gems/json-1.8.2/lib/json/common.rb:468:in `JSON'
```

I figured out that this error happens because I use `oj` as a json parser. 
I have in Gemfile these two gems

```
#https://github.com/ohler55/oj
gem 'oj'
#https://github.com/ohler55/oj_mimic_json
gem 'oj_mimic_json' #required for Rails 4.1.x
```

If I remove `oj_mimic_json` everything works fine

What do you think about this? 
Thank you

claudiofullscreen:
Thanks @MaximSt – now I understand the error.

The line `gem 'oj_mimic_json'` tells Ruby to use OJ whenever JSON is in the code. See [this file](https://github.com/ohler55/oj_mimic_json/blob/master/lib/oj_mimic_json.rb) which calls `Oj.mimic_JSON()`.

However, [as documented in oj](https://github.com/ohler55/oj/blob/master/ext/oj/oj.c#L1840L1846): 

> **Most** options that could be passed to the JSON methods are supported.

Most, but not all of them. In fact, if you look at the `parse` in Oj [source code](https://github.com/ohler55/oj/blob/master/ext/oj/oj.c#L2253L2265) you will see that **it's not compatible with JSON.parse** because it has a different signature.

My advice: don't use `oj_mimic_json`. Use `oj` and then call `OJ.parse` when you want to parse a JSON file with that library, and use `JSON.parse` when you want to parse a JSON file with the Ruby JSON library (as Yt does).

maksymst:
Thank you @claudiofullscreen

Issue #173, opened 2015-05-15
Is there any age restriction for the video


praneeththota:
Is there any method to know age restriction for the video. I tried like this
video = Yt::Video.new url: 'https://www.youtube.com/watch?v=p5Lj3y3mOxw'
video.annotation.above? 50
error: NoMethodError: undefined method `annotation' for #Yt::Models::Video:0x007f30d62b34b0

video.annotations.above? 50
error: NoMethodError: undefined method `annotation' for #Yt::Models::Video:0x007f30d62b34b0

praneeththota:
I heard that YouTube API V3 does not provide access to video annotations, therefore a legacy XML endpoint is used to retrieve annotations.
How to get the XML endpoint to retrieve annotations in rails ?

claudiofullscreen:
Hello @praneeththota.

Regarding the error you get, there is no `.annotation` (singular) method on `Yt::Video`; instead there is a plural `.annotations` method that returns all the annotations [as documented here](http://fullscreen.github.io/yt/annotations.html)

Therefore you could write `video.annotations.first.above? 50` to know if the first annotation of the video is in the top half of the video frame.

However, that is not related to "age restriction". I looked at the YouTube API V3 documentation and this is what I found:

![screen shot 2015-05-15 at 7 12 03 am](https://cloud.githubusercontent.com/assets/7408595/7654414/be5b1230-fad1-11e4-9c79-1cca5c6af9c2.png)

If this is what you are looking for, it's not implemented in Yt… but pull requests are welcome!

jacknguyen:
@praneeththota like @claudiob said there is no single method to check for content rating of a video but you can do so by checking the content_detail of the video and parsing down until you find the contentRating.

What I was able to do:
`video.content_detail.data['contentRating']`

Youtube can send back a number of possible values back which is [documented here](https://developers.google.com/youtube/v3/docs/videos) under `contentDetails.contentRating`. But I think the most common one might just be `ytRating` which only has only one possible return value `ytAgeRestricted`.

praneeththota:
Yeah exactly i'm searching for this. Thanks alot @jacknguyen , @claudiofullscreen

PR #197, opened 2015-06-09
Method to check if a video has any age restriction.

jacknguyen:


lib/yt/models/video.rb
```diff
@@ -247,6 +247,12 @@ def licensed?
         content_detail.licensed_content || false
       end
 
+      # @return [Boolean] whether the video is identified by YouTube as
+      #   age-restricted content.
+      def age_restricted?
+        content_detail.content_rating != nil
```
claudiofullscreen: Hello @jacknguyen and thanks for your contribution! 
The source code of yt is a little hard to grasp at first, and you went right for it! :clap: 

Regarding this line, you are stating that having _any_ value in content rating means that the video is "age-restricted". 

However there are [many different values according to the docs](https://developers.google.com/youtube/v3/docs/videos#contentDetails.contentRating).
For instance, the value might indicate that:

> These programs are intended for children (other than preschool children) who are younger than 14 years of age.

In that case, the video is not really "restricted" to an age, just recommended.
Could you elaborate on a case where you needed to know if there was _any_ content-rating in place, independently of the value of the content rating itself?

claudiofullscreen: Looking back at your https://github.com/Fullscreen/yt/issues/173#issuecomment-104075497 I realize that maybe what you intended to go for is to check if there is any value for `contentDetails.contentRating.ytRating` instead.

That would make sense. If `"contentRating"` is a Hash and the `"ytRating"` key as the `"ytAgeRestricted"` value, then the video is age-restricted. If you change your pull request to check for that instead, I'll accept it :+1: 

If possible, squash your commits too :dancers:

jacknguyen:
Hi @claudiofullscreen.

That was actually also a concern I had with only checking whether contentRating was nil.

I had to include the check for nil as well as ytRating because I kept getting no method error when trying to access the Hash of a nilClass. Made the line a little longer than I would like.

Also I was able to squash the commits, I hope you didn't see my earlier attempts at it :smile:.

jacknguyen:
I'll also update the documentation as well, once you accept. :+1:

claudiofullscreen:
@jacknguyen I've accepted your commit in #198 which also adds https://github.com/Fullscreen/yt/commit/69906de4f98eea30558bb8d7f75cf26d8df4d9a7 to slightly reduce the amount of code and add the CHANGELOG.

Feel free to open a Pull Request against the `gh-pages` branch now if you'd like to update fullscreen.github.io/yt accordingly :grin: 

Thanks for your work!

PR #217, opened 2015-07-14
Added ContentOwnerAdvertisingOptions

coryodaniel:
The names all mimic the youtube "kind" key in the resource hashes. The names are pretty lengthy


lib/yt/actions/base.rb
```diff
@@ -29,4 +29,4 @@ def sanitize_brackets!(source)
       end
     end
   end
-end
\ No newline at end of file
+end
```
claudiofullscreen, line 32: You are right that there should be a new line here. 

However… either make it a separate commit or don't add it here.

It's just confusing to see this file touched as part of this PR.

Thanks.

lib/yt/collections/content_owner_advertising_options_sets.rb
```diff
@@ -0,0 +1,34 @@
+require 'yt/collections/resources'
+require 'yt/models/content_owner_advertising_options_set'
+
+module Yt
+  module Collections
+    # Provides methods to interact with a the content owner advertising options of a YouTube video.
+    #
+    # Resources with advertising options are: {Yt::Models::ContentOwner}.
+    class ContentOwnerAdvertisingOptionsSets < Resources
```
claudiofullscreen, line 9: For as long as this class name is, it's ok to stick with it for now.

PR #223, opened 2015-07-27
Add `related_playlist` to Account and Channel

### LINKED ISSUE #218: Where params aren't respected for playlists
I believe we'd need something along these lines: https://github.com/Fullscreen/yt/blob/master/lib/yt/collections/channels.rb#L24-L28

claudiofullscreen:
With related_playlists, you can retrieve special playlists associated
to a channel such as "Liked videos" or "Uploads"

Closes #218

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
+## 0.25.4 - 2015-07-27
+
+* [FEATURE] Add `channel.related_playlist` and `account.related_playlists` to access "Liked Videos", "Uploads", etc.
+
 ## 0.25.3 - 2015-07-23
 
 * [BUGFIX] Don’t run an infinite loop when calling `.playlist_items.includes(:video)` on a playlist with only private or deleted videos
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ To install on your system, run
 
 To use inside a bundled Ruby project, add this line to the Gemfile:
 
-    gem 'yt', '~> 0.25.3'
+    gem 'yt', '~> 0.25.4'
 
 Since the gem follows [Semantic Versioning](http://semver.org),
 indicating the full version in your Gemfile (~> *major*.*minor*.*patch*)
--- a/YOUTUBE_IT.md
+++ b/YOUTUBE_IT.md
@@ -180,37 +180,18 @@ channel = Yt::Channel.new url: 'youtube.com/liz'
 channel.videos.where(q: 'penguin')
 ```
 
-<!--
-
-TODO: Although FAVORITE is somehow deprecated, it can still be retrieved with
-V3. First, get the contentDetails of the account's channel and you'll get
-
-{"relatedPlaylists"=>
-  {"likes"=>"LLrWYT-_Xncr1BLPfPNr50sQ",
-   "favorites"=>"FLrWYT-_Xncr1BLPfPNr50sQ",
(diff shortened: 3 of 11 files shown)


claudiofullscreen:
@trevorturk This should work to close #218 

I need to warn you that YouTube has deprecated "Favorites". 
Therefore _old_ YouTube channels will have a related playlist called "Favorites", but new accounts won’t.
Every account should have the "Liked videos" playlist, though.

trevorturk:
This looks great! Would you like me to test things out now, or should I wait until you merge this into the master branch? 

Thanks for the warning about the deprecated "Favorites" playlist. Actually, I'm finding that things are quite odd even with new accounts. For example, here's a new test account that I made, which has 3 different playlists returned by the API with various calls:

First, the "Favorites" playlist for user (returned by relatedPlaylists) does not exist when requested via the website when logged in as the user: https://www.youtube.com/playlist?list=FLIbyQMQMIm3HAbJ1IXaOFaQ

Second, the "Liked videos" playlist for user (returned by relatedPlaylists) is private: https://www.youtube.com/playlist?list=LLIbyQMQMIm3HAbJ1IXaOFaQ 

Third, this oddball "Favorites" playlist for the channel (returned as the only playlist for the channel when requested via the API) seems to combine the above "Favorites" and "Liked videos" even though the "Favorites" playlist from relatedPlaylists allegedly doesn't exist. (Videos appear in this playlist if you use the "Add to" menu and select the allegedly nonexistent "Favorites" playlist _or_ if click the "Like" / thumbs-up button on any video: https://www.youtube.com/playlist?list=PLY84RbNn14ftXWknvOGm95Z6qjkJi-Qpl 

In my use case, I want to collect all favorites/likes for a given user. So, with older accounts I'm able to use the single relatedPlaylists call to get the (valid) playlist ids for "Favorites" and "Liked videos" and then I can collect the videos from those two lists. For newer accounts that have two different "Favorites" playlist ids (one gotten from relatedPlaylists and one gotten from the channel's playlists) I collect videos from the channel's "Favorites" playlist, which mysteriously combines the two separate playlists into a single combined playlist. 

Sorry, I know that's confusing, but anyway I thought it worth trying to explain :)

claudiofullscreen:
Hello @trevorturk !

My first suggestion is to stop using "Favorites" at all.
YouTube deprecated "Favorites". They might still be around for old users, but they will be gone soon. 

I don't seem to find the blog post where they announced it, but [this link](https://developers.google.com/youtube/analytics/v1/dimsmets/mets#favoritesAdded) should work.

I also noticed that, when you ask for your own "related playlists", YouTube includes some IDs that do not exist at all. I believe this is a YouTube bug and I will report this to YouTube.

However, I have already taken care of this weirdness in this PR by eager-loading the `Yt::Playlist`, not just their IDs. Therefore, I invite you to try your code against this branch. You should see that those "unknown" playlists **are not included** in `account.related_playlists` at all.

Regarding the fact that the "Liked Videos" playlist for a user is private, there's nothing I can do about it. Some users have this playlist set to public (such as [this one](https://www.youtube.com/playlist?list=LLxO1tY8h1AhOz0T4ENwmpow)), other users have set it to private and… looks like YouTube does not let you change the privacy status of this playlist anymore!

Summing up, I think this PR might still help you retrieve _some data_ for _some YouTube users_. Better than nothing, but it won't solve YouTube bugs or weirdnesses. Give it a try and let me know.

trevorturk:
That's great about the eager-loading. I'll give this branch a try now and report back.

trevorturk:
I just tested with my small sample set of accounts and it looks great! Thank you!

claudiofullscreen:
@trevorturk Update to `0.25.4` and you should be good to go!

trevorturk:
Awesome! Thank you <3

PR #240, opened 2015-08-31
Make Collections::Base fully enumerable

topalovic:
Stumbled upon `Yt::Collections::Base` not having `#reject`, but actually having `#select`. Same with `#map` but no `#collect` :unamused:

I believe it's reasonable to expect a fully `Enumerable` base collection. I checked the pulls and it seems to be a common request, with folks adding methods piecemeal.

I don't have any particular tests in mind for this. I'm sure I can provide some if you think it's worthwhile.

### DIFF
--- a/lib/yt/actions/list.rb
+++ b/lib/yt/actions/list.rb
@@ -6,8 +6,7 @@ require 'yt/config'
 module Yt
   module Actions
     module List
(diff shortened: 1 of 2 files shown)


claudiofullscreen:
This is possibly a duplicate of https://github.com/Fullscreen/yt/pull/119

PR #250, opened 2015-11-05
handle youtube.com/v/<id> urls

andrewroth:
handle youtube.com/v/<id> urls

### DIFF
--- a/lib/yt/models/url.rb
+++ b/lib/yt/models/url.rb
@@ -72,6 +72,7 @@ module Yt
           youtube\\.com/watch\\?v=#{video_id}
           youtu\\.be/#{video_id}
           youtube\\.com/embed/#{video_id}
+          youtube\\.com/v/#{video_id}
         }
       end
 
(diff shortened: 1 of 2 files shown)


claudiofullscreen:
Wow, I didn't know that was a thing.

And actually… I'm going to have to update to youtube.com/c/... for channels too :sob:

claudiofullscreen:
Thanks!! :+1: :tada: :grin:

stavro:
FYI `youtube.com/c/` urls are Google+ names, not YouTube usernames, they can change (eg http://youtube.com/c/sÉanswild), and they were not exposed via a lookup API last time I checked.. so I don't believe `Yt` can support them right now :cry:

andrewroth:
No prob.. I was looking for a list of all the different ways of accessing a yt video, but I couldn't find anything.. It seems /v/ is some sort of full-screen view.. we have some users on our site entering youtube urls like that

PR #235, opened 2015-08-07
Add comments resource and link it with videos

ezr-ondrej:
Comments resource support

PR #269, opened 2016-03-22
Add Yt::CommentThread model and related tests

MoTai:
This is an implementation of YouTube comment thread resource:

https://developers.google.com/youtube/v3/docs/commentThreads

### DIFF
--- a/lib/yt/models/snippet.rb
+++ b/lib/yt/models/snippet.rb
@@ -8,6 +8,7 @@ module Yt
     # @see https://developers.google.com/youtube/v3/docs/videos#resource
     # @see https://developers.google.com/youtube/v3/docs/playlists#resource
     # @see https://developers.google.com/youtube/v3/docs/playlistItems#resource
+    # @see https://developers.google.com/youtube/v3/docs/commentThread#resource
     class Snippet < Base
       attr_reader :data
 
@@ -28,9 +29,19 @@ module Yt
       has_attribute :position, type: Integer
       has_attribute :resource_id, default: {}
       has_attribute :thumbnails, default: {}
+      has_attribute :video_id
+      has_attribute :total_reply_count, type: Integer
 
       def thumbnail_url(size = :default)
         thumbnails.fetch(size.to_s, {})['url']
+      end
+
+      def public?
+        @public ||= data.fetch 'isPublic', false
+      end
+
+      def can_reply?
+        @can_reply ||= data.fetch 'canReply', false
       end
 
       # Returns whether YouTube API includes all attributes in this snippet.
--- a/spec/requests/as_server_app/comment_thread_spec.rb
+++ b/spec/requests/as_server_app/comment_thread_spec.rb
@@ -1,0 +1,26 @@
+require 'spec_helper'
+require 'yt/models/comment_thread'
+
+describe Yt::CommentThread, :server_app do
+  subject(:comment_thread) { Yt::CommentThread.new attrs }
+
+  context 'given an existing comment thread ID about a channel' do
+    let(:attrs) { {id: 'z13vsnnbwtv4sbnug232erczcmi3wzaug'} }
+
+    it { expect(comment_thread.video_id).to be_nil }
+    it { expect(comment_thread.total_reply_count).to be_an Integer }
+    it { expect(comment_thread.can_reply?).to be false }
+    it { expect(comment_thread).to be_public }
+  end
+
+  context 'given an comment thread ID about a video' do
+    let(:attrs) { {id: 'z13ij10h2z3qxpcte23hc5oh2vfzeptk4'} }
+    it { expect(comment_thread.video_id).to be_a String }
+  end
+
+  context 'given an unknown comment thread ID' do
+    let(:attrs) { {id: 'not-a-comment-thread-id'} }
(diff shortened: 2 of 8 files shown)


claudiofullscreen:
Taking a look, thanks!!! :+1:

spec/requests/as_server_app/comment_thread_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+require 'yt/models/comment_thread'
+
+describe Yt::CommentThread, :server_app do
+  subject(:comment_thread) { Yt::CommentThread.new attrs }
+
+  context 'given an existing comment thread ID about a channel' do
+    let(:attrs) { {id: 'z13vsnnbwtv4sbnug232erczcmi3wzaug'} }
+
+    it { expect(comment_thread.video_id).to be_nil }
+    # Not sure why but channel_id seems always missing value
```
claudiofullscreen: Typical of YouTube to have wrong documentation.

Since you probably don't need this for your app, just remove it completely from code and tests.

spec/support/boolean_matcher.rb
```diff
@@ -0,0 +1,5 @@
+RSpec::Matchers.define :be_boolean do
+  match do |actual|
+    !!actual == actual
+  end
+end
```
claudiofullscreen: We actually want boolean methods, not fake ones.

For instance look at https://github.com/Fullscreen/yt/blob/master/lib/yt/models/snippet.rb#L44-L47

lib/yt/models/snippet.rb
```diff
@@ -29,6 +29,11 @@ def initialize(options = {})
       has_attribute :resource_id, default: {}
       has_attribute :thumbnails, default: {}
 
+      has_attribute :video_id
+      has_attribute :total_reply_count, default: 0
+      has_attribute :is_public, default: true
+      has_attribute :can_reply, default: true
```
claudiofullscreen: Look at https://github.com/Fullscreen/yt/blob/master/lib/yt/models/snippet.rb#L44-L47

We want to transform the response of YouTube into a (Ruby) boolean.

spec/requests/as_server_app/comment_thread_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'spec_helper'
+require 'yt/models/comment_thread'
+
+describe Yt::CommentThread, :server_app do
+  subject(:comment_thread) { Yt::CommentThread.new attrs }
+
+  context 'given an existing comment thread ID about a channel' do
+    let(:attrs) { {id: 'z13vsnnbwtv4sbnug232erczcmi3wzaug'} }
+
+    it { expect(comment_thread.video_id).to be_nil }
```
claudiofullscreen, line 10: Why is `video_id` nil? Shouldn't it exist?

spec/requests/as_server_app/comment_thread_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    # Not sure why but channel_id seems always missing value
+    it { expect(comment_thread.channel_id).to be_nil }
+    it { expect(comment_thread.total_reply_count).to be_an Integer }
+    it { expect(comment_thread.can_reply).to be_boolean }
+    it { expect(comment_thread.is_public).to be_boolean }
```
claudiofullscreen: Once you have _real_ booleans, you can write tests like this:

https://github.com/Fullscreen/yt/blob/master/spec/models/resource_spec.rb#L22-L26

spec/requests/as_server_app/comment_thread_spec.rb
claudiofullscreen, line 10: Oh, I see the example below has a video ID. 

So I think I understand what is going on here.

YouTube allows people to comment **on** a video and (separately) **on** a channel.

So probably video_id is present in one case (and channel_id missing) and vice versa.

For instance, these are "Channel comments": https://www.youtube.com/user/Fullscreen/discussion

claudiofullscreen:
@MoTai Good job. Check out my code reviews :grin:

spec/requests/as_server_app/comment_thread_spec.rb
MoTai: I am thinking maybe the following?

``` ruby
# lib/models/snippet.rb
has_attribute :is_public, default: true
```

``` ruby
# lib/models/comment_thread.rb
delegate :is_public, to: :snippet

def is_public?
  is_public
end
```

``` ruby
# tests
it { expect(comment_thread.is_public?).to be true }
```

Cheers,

claudiofullscreen: I was suggesting something like this

``` ruby
# lib/models/snippet
def public?
  @public ||= data.fetch :is_public, false
end
```

``` ruby
# lib/models/comment_thread.rb
delegate :public?, to: :snippet
```

``` ruby
# tests
it { expect(comment_thread).to be_public }
```

claudiofullscreen:
This is similar to #235 but with tests :clap:

claudiofullscreen:
Thank you @MoTai !!!! :ribbon:

MoTai:
:laughing:

PR #272, opened 2016-04-05
Asset search

philipqnguyen:
Currently, if asset ID was never set when instantiating `Yt::ContentOwner`, calling #assets would still use the path,  /youtube/partner/v1/assets. Unfortunately that would not work because that endpoint expects an asset ID. So this should fix where if the asset ID was never set, it will use the path, /youtube/partner/v1/assetSearch, otherwise it will continue to use the path, /youtube/partner/v1/assets.


claudiofullscreen:
Great job!!! :clap: :clap: :clap: :clap: :clap: :bow: :tada: :tada: :tada: :tada: :tada:

PR #274, opened 2016-04-07
Youtube has more than one types of rate limits, so 'Exceeded' should cover more.

philipqnguyen:
Previously, "quotaExceeded" was too specific, so "Exceeded" should cover more of youtube's rate limit errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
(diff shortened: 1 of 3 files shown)


claudiofullscreen:
Great!

I would be even more generic and make that reggae _case insensitive_.

Then add CHANGELOG, bump version :)

claudiofullscreen:
Thanks OS X for auto-correcting regex into reggae

philipqnguyen:
Pushed up some more commits. Should be case insensitive now. Version bumped to 0.25.29 and added to the changelog.

claudiofullscreen:
Great job! :clap: :dancer: :tada:

PR #277, opened 2016-04-12
Fix: remove has_one :claim from Yt::Video

claudiofullscreen:
We cannot assume that every video has a claim. We can still try to
eagerly loaded the claims when we need to, but we should not use
`has_one`, since it assumes the existence of the association.


claudiofullscreen:
@philipqnguyen This should fix your code :wind_chime:

PR #278, opened 2016-04-12
Keep :claim, :category in included_relationships, ignore it if neccessary

philipqnguyen:
So a problem we had was that claims and video_category were only being eager loaded for the first videos API call. Subsequent lazily loaded videos did not eager load the associated claims or categories. It happened because we deleted the `:video` and `:category` to prevent it from being included in `parts`, but it was never added back in.

So we keep `:claim` and `:category` in the the array, `included_relationships`. In `parts`, we reject the `:claim` and `:category` so that it doesn't get passed into the parts params.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
-## 0.25.32 - 2016-04-11
+## 0.25.32 - 2016-04-12
 
 * [BUGFIX] Fix where videos did not eager load claims or categories in subsequent requests.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
+## 0.25.32 - 2016-04-11
+
+* [BUGFIX] Fix where videos did not eager load claims or categories in subsequent requests.
+
 ## 0.25.31 - 2016-04-11
 
 * [BUGFIX] Don’t try to instantiate video.claim if a video does not have a claim.
--- a/lib/yt/collections/videos.rb
+++ b/lib/yt/collections/videos.rb
@@ -33,13 +33,15 @@ module Yt
 
       def eager_load_items_from(items)
         if included_relationships.any?
-          include_category = included_relationships.delete(:category)
-          include_claim = included_relationships.delete(:claim)
-          included_relationships.append(:snippet).uniq! if include_category || include_claim
+          included_relationships.append(:snippet).uniq! if
+            included_relationships.include?(:claim) ||
+            included_relationships.include?(:category)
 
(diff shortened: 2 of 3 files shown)


lib/yt/collections/videos.rb
```diff
@@ -33,13 +33,15 @@ def attributes_for_new_item(data)
 
       def eager_load_items_from(items)
         if included_relationships.any?
-          include_category = included_relationships.delete(:category)
-          include_claim = included_relationships.delete(:claim)
-          included_relationships.append(:snippet).uniq! if include_category || include_claim
+          included_relationships.append(:snippet).uniq! if
+            included_relationships.include?(:claim) ||
+            included_relationships.include?(:category)
```
claudiofullscreen: How about

``` ruby
associations = [:claim, :category]

if (included_relationships & associations).any?
  included_relationships.append(:snippet).uniq!
end
```

lib/yt/collections/videos.rb
```diff
@@ -33,13 +33,15 @@ def attributes_for_new_item(data)
...
 
           ids = items.map{|item| item['id']['videoId']}
-          parts = included_relationships.map{|r| r.to_s.camelize(:lower)}
-          conditions = {id: ids.join(','), part: parts.join(',')}
+          parts = included_relationships.reject do |p|
+            p == :claim || p == :category
+          end.map{|r| r.to_s.camelize(:lower)}
+          conditions = { id: ids.join(','), part: parts.join(',') }
```
claudiofullscreen, line 45: and here 

``` ruby
parts = (included_relationships - associations).map{|r| r.to_s.camelize(:lower)}
```

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
+## 0.25.32 - 2016-04-11
```
claudiofullscreen: 2016-04-12 :grin:

claudiofullscreen:
@philipqnguyen check my comments. other than that, looks :ok:

MoTai:
👍👏  This caused me some serious headache when debugging the issue.

philipqnguyen:
@claudiofullscreen I made the suggested changes =D

lib/yt/collections/videos.rb
```diff
@@ -33,13 +33,16 @@ def attributes_for_new_item(data)
...
+            included_relationships.append(:snippet).uniq!
+          end
 
           ids = items.map{|item| item['id']['videoId']}
-          parts = included_relationships.map{|r| r.to_s.camelize(:lower)}
-          conditions = {id: ids.join(','), part: parts.join(',')}
+          parts = (included_relationships - associations).map do |r|
+            r.to_s.camelize(:lower)
+          end
+          conditions = { id: ids.join(','), part: parts.join(',') }
```
claudiofullscreen, line 45: you really like those extra spaces :grimacing:

philipqnguyen, line 45: I used to have rubocop hooked up in some other projects and by default it demanded me to put in those spaces between curly braces lol.

PR #279, opened 2016-04-15
Video eagerload

philipqnguyen:
These changes add:
- `$content_owner.videos.includes(:claim)` will so eager-load assets. Allowing you to do this: `$content_owner.videos.includes(:claim).first.claim.asset.id`.
- Add `Claim#source` that returns the source of the claim.

### DIFF
--- a/lib/yt/models/claim.rb
+++ b/lib/yt/models/claim.rb
@@ -114,12 +114,12 @@ module Yt
         content_type == 'audiovisual'
       end
 
-      # @return [Yt::Models::Claim] the source of the claim
+      # @return [String] the source of the claim
       def source
         @data.fetch('origin', {})['source']
       end
 
-      # @return [Yt::Models::Claim] the asset of the claim
+      # @return [Yt::Models::Asset] the asset of the claim
       def asset
         @asset
       end
--- a/lib/yt/models/claim.rb
+++ b/lib/yt/models/claim.rb
@@ -11,6 +11,7 @@ module Yt
         @data = options[:data]
         @id = options[:id]
         @auth = options[:auth]
+        @asset = options[:asset] if options[:asset]
       end
 
       # @!attribute [r] claim_history
@@ -117,6 +118,11 @@ module Yt
       def source
         origin = @data.fetch 'origin', nil
         origin.fetch 'source', nil if origin
+      end
+
+      # @return the first asset of the claim.
+      def asset
+        @asset
       end
 
       # @return [Time] the date and time that the claim was created.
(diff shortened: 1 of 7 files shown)


spec/requests/as_content_owner/content_owner_spec.rb
```diff
@@ -53,19 +53,15 @@
     describe '.includes(:claim)' do
       let(:videos) { $content_owner.videos.includes(:claim) }
       let(:video_with_claim) { videos.find{|v| v.claim.present?} }
+      let(:video_with_claim_with_asset) do
+        videos.find{|v| v.try(:claim).try(:asset)}
+      end
```
claudiofullscreen: I think we can get rid of this, and just use the previous `video_with_claim`. 

In other words, every video with claim will also have an asset, and we just expect it to be eager-loaded too.

spec/requests/as_content_owner/content_owner_spec.rb
```diff
@@ -53,19 +53,15 @@
...
 
       specify 'eager-loads the claim of each video' do
```
claudiofullscreen: and change this to

``` ruby
specify 'eager-loads the claim of each video and its asset' do
  expect(video_with_claim.claim).to be_a Yt::Claim
  expect(video_with_claim.claim.id).to be_a String
  expect(video_with_claim.claim.video_id).to eq video_with_claim.id

  expect(video_with_claim.claim.asset).to be_a Yt::Asset
  expect(video_with_claim.claim.asset.id).to be_a String 
end
```

lib/yt/models/claim.rb
```diff
@@ -114,13 +114,12 @@ def audiovisual?
...
-      # @return String of the claim's source.
+      # @return [Yt::Models::Claim] the source of the claim
       def source
-        origin = @data.fetch 'origin', nil
-        origin.fetch 'source', nil if origin
+        @data.fetch('origin', {})['source']
       end
 
-      # @return the first asset of the claim.
+      # @return [Yt::Models::Claim] the asset of the claim
```
claudiofullscreen: Is this `Yt::Models::Claim` or `[Yt::Models::Asset]`?

lib/yt/models/claim.rb
```diff
@@ -114,13 +114,12 @@ def audiovisual?
         content_type == 'audiovisual'
       end
 
-      # @return String of the claim's source.
+      # @return [Yt::Models::Claim] the source of the claim
```
claudiofullscreen: I think the source is actually a string, such as `webUpload`.

lib/yt/models/claim.rb
philipqnguyen: Yes, it would be asset. Fixing now =D

lib/yt/models/claim.rb
philipqnguyen: Yes it is

PR #282, opened 2016-05-10
Update conditions to not raise error in authentications

kangkyu:
to not retry authentication code in case "already redeemed" or "invalid", involving the error description value in `expected?`

```
{
  "error" : "invalid_grant",
  "error_description" : "Code was already redeemed."
}
```

```
{
  "error" : "invalid_grant",
  "error_description" : "Invalid code."
}
```

And make private methods `explanation` and `response_body` under RequestError into public

### DIFF
--- a/lib/yt/collections/authentications.rb
+++ b/lib/yt/collections/authentications.rb
@@ -40,7 +40,14 @@ module Yt
       end
 
       def expected?(error)
-        error.kind == 'invalid_grant'
+        error.kind == 'invalid_grant' &&
+          invalid_code_errors.exclude?(error.description)
+      end
+
+    private
+
+      def invalid_code_errors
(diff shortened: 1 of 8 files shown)


lib/yt/collections/authentications.rb
```diff
@@ -40,8 +40,9 @@ def next_page
       end
 
       def expected?(error)
-        error.kind == 'invalid_grant'
+        error.kind == 'invalid_grant' &&
+        !(error.description == "Code was already redeemed." || error.description == "Invalid code.")
```
claudiofullscreen: Can you make a variable or method that returns `["Code was already redeemed.", "Invalid code."]`?

Then the code will be shorter (you can use `in?` or `include?`)

PR #297, opened 2016-10-06
Update gem to 0.28.0

bruce-fs:
- Change earnings & impressions to estimated_revenue and ad_impressions
- source:
  https://developers.google.com/youtube/analytics/revision_history#august-10-2016

PR #301, opened 2016-10-17
Make requests with new metric names

kangkyu:
Change names: earnings -> estimated_revenue, impressions -> add_impressions
Add deprecation message for old metric names

### DIFF
--- a/lib/yt/associations/has_reports.rb
+++ b/lib/yt/associations/has_reports.rb
@@ -225,13 +225,6 @@ module Yt
 
     private
 
-      def deprecated_alias(deprecated, replacement)
-        define_method deprecated do |options = {}|
-          warn "##{deprecated} has been deprecated. please use ##{replacement}"
-          send replacement, options
-        end
-      end
-
       def define_reports_method(metric, type)
         (@metrics ||= {})[metric] = type
         define_method :reports do |options = {}|
@@ -306,7 +299,7 @@ module Yt
 
       def define_all_metric_method(metric, type)
         define_method "all_#{metric}" do
-          # @note Asking for the "earnings" metric of a day in which a channel
+          # @note Asking for the "estimated_revenue" metric of a day in which a channel
           # made 0 USD returns the wrong "nil". But adding to the request the
           # "estimatedMinutesWatched" metric returns the correct value 0.
           metrics = {metric => type}
--- a/lib/yt/associations/has_reports.rb
+++ b/lib/yt/associations/has_reports.rb
@@ -225,6 +225,13 @@ module Yt
 
     private
 
+      def deprecated_alias(deprecated, replacement)
+        define_method deprecated do |options = {}|
+          warn "##{deprecated} has been deprecated. please use ##{replacement}"
+          send replacement, options
+        end
+      end
+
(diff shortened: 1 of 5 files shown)


lib/yt/collections/reports.rb
```diff
@@ -179,8 +179,8 @@ def reports_params
           params['max-results'] = 50 if @dimension.in? [:playlist, :video]
           params['max-results'] = 25 if @dimension.in? [:embedded_player_location, :related_video, :search_term, :referrer]
           if @dimension.in? [:video, :playlist, :embedded_player_location, :related_video, :search_term, :referrer]
-            if @metrics.keys == [:earnings, :estimated_minutes_watched]
-              params['sort'] = '-earnings'
+            if @metrics.keys == [:estimated_revenue, :estimated_minutes_watched]
+              params['sort'] = '-estimated_revenue'
```
claudiofullscreen: @kangkyu Does this work? Or should it be `-estimatedRevenue`?

claudiofullscreen:
Closes #297

lib/yt/associations/has_reports.rb
```diff
@@ -225,6 +225,13 @@ def has_report(metric, type)
 
     private
 
+      def deprecated_alias(deprecated, replacement)
+        define_method deprecated do |options = {}|
+          warn "##{deprecated} has been deprecated. please use ##{replacement}"
+          send replacement, options
+        end
+      end
```
claudiofullscreen: @kangkyu this is a great idea.

However, let's not deprecate those methods in this PR, since those are methods that we are using in the tests and in the documentation.

If we decide to deprecate them, then we'll have to change those as well.

For now you can just `alias` or `alias_method`

lib/yt/collections/reports.rb
```diff
@@ -136,7 +136,7 @@ def within(days_range, country, state, dimension, videos, try_again = true)
           elsif dimension.in? [:traffic_source, :country, :state, :playback_location, :device_type, :operating_system, :subscribed_status]
             hash = hash.transform_values{|h| h.sort_by{|range, v| -v}.to_h}
           end
-          (@metrics.one? || @metrics.keys == [:earnings, :estimated_minutes_watched]) ? hash[@metrics.keys.first] : hash
+          (@metrics.one? || @metrics.keys == [:estimated_revenue, :estimated_minutes_watched]) ? hash[@metrics.keys.first] : hash
```
claudiofullscreen, line 139: @kangkyu apart from testing calls like `video.earnings`, did you also try `video.reports(only: [:earnings])`? would that type of call also work?

kangkyu:
``` rb
channel.reports :only=>[:earnings], :total=>true
# this was not working
channel.reports :only=>[:estimated_revenue], :total=>true
# this was working
channel.estimated_revenue
# working
channel.earnings
# working
```

I tried these on `bundle console` with YT_TEST values

kangkyu:
Aliasing wasn't enough for keeping deprecated metric names  (`channel.earnings` working but `channel.reports only: [:earnings]` not working) so close this pull request in favor of #297

PR #326, opened 2017-07-03
Add include-historical-channel-data by option historical: true

kangkyu:
By this change, include `include-historical-channel-data=true` or `include-historical-channel-data=false` option into the query only when `:historical` option being passed

### DIFF
--- a/lib/yt/collections/reports.rb
+++ b/lib/yt/collections/reports.rb
@@ -117,12 +117,13 @@ module Yt
 
       attr_writer :metrics
 
-      def within(days_range, country, state, dimension, videos, max_retries = 3)
+      def within(days_range, country, state, dimension, videos, historical, max_retries = 3)
         @days_range = days_range
         @country = country
         @state = state
         @dimension = dimension
         @videos = videos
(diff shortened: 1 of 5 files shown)


lib/yt/collections/reports.rb
```diff
@@ -188,7 +189,7 @@ def reports_params
           params['end-date'] = @days_range.end
           params['metrics'] = @metrics.keys.join(',').to_s.camelize(:lower)
           params['dimensions'] = DIMENSIONS[@dimension][:name] unless @dimension == :range
-          params['include-historical-channel-data'] = 'true'
+          params.merge!({'include-historical-channel-data' => @historical}) if @historical
```
claudiofullscreen: @kangkyu Follow the other lines and avoid `merge!` since this might have side-effects if other methods call this code again:

```ruby
params['include-historical-channel-data'] = @historical if @historical
```

claudiofullscreen:
@kangkyu apart from that comment, version (patch), changelog, explain how to upgrade (readme), release!

PR #341, opened 2019-02-26
Add update method to Yt::Claim

jcohenho:
* Adds method to update Yt::Claim


kangkyu:
thank you @jcohenho for your patience we are planning to review pull requests soon

jcohenho:
Hello, 

When will PR Reviews happen?

efremroberson (approved):
@8bitDesigner says hi

spec/requests/as_content_owner/claim_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+describe Yt::Claim, :partner do
+  describe '.ownership' do
+    let(:claim) { Yt::Claim.new id: claim_id, auth: $content_owner }
+
+    describe 'given a claim administered by the content owner' do
+      let(:claim_id) { ENV['YT_TEST_PARTNER_CLAIM_ID'] }
+
+      describe 'the claim can be updated' do
+        let(:attrs) { {status: 'inactive'} }
```
kangkyu: To make it inactive, it has `delete` method already. I think we need more specific case

lib/yt/models/claim.rb
```diff
@@ -14,6 +14,12 @@ def initialize(options = {})
         @asset = options[:asset] if options[:asset]
       end
 
+      def update(attributes = {})
```
kangkyu, line 22: Will you add some method documentation on top for a public method? that would be nice

then I think this part should be included, to avoid surprises
> If you are submitting an update request, and your request does not specify a value for a property that already has a value, the property's existing value will be deleted.

dgb:
This LGTM once feedback is addressed. Thanks for bearing with us @jcohenho as we ramp up maintaining this gem.

lib/yt/models/claim.rb
jcohenho: updated

spec/requests/as_content_owner/claim_spec.rb
jcohenho: updated

jcohenho:
Updated as per feedback

dgb:
Wanna try merging master in? I think that's why the build is failing.

jcohenho:
Looks like other tests are failing already

dgb:
@jcohenho yeah, looks like we're getting rate limited. I guess I need to work on getting the test suite vcr-d sooner than I thought. I'll try rerunning the suite on Monday. Thanks again for bearing with us.

jcohenho:
@dgb still seeing errors in the tests

dgb:
I remembered that Travis strips the environment variables from non-member PRs, so that's why the tests are failing 🙈 . I'll set aside some time today to run the tests locally and merge/update as necessary. Thanks.

PR #342, opened 2019-02-26
Add method to Yt::Asset to retrieve asset metadata

jcohenho:
Add method to Yt::Asset to retrieve asset metadata

Add  method to Yt::Asset to retrieve asset metadata

Revert changes to Yt::Claim

Revert old changes

Add default to asset label attribute


jcohenho:
Hello, 

When will PR Reviews happen?

dgb:
Hi there,

My name is Derek and I'm now one of the maintainers of the Yt gem. I want to both thank you for your contribution and apologize for the lack of feedback on your pull request. Maintaining an open source project like Yt is a pretty big job, and we're taking steps to prioritize maintenance and development of the project. We're hoping to review community contributions on a regular basis, as well as provide a roadmap for our development plans.

As part of this process we're going through the open PRs and taking action on each one, and for this PR we've chosen to review it. You can expect some more specific feedback in the next week or so.

Thanks again for your contribution, and please let us know if you have any questions or comments.

lib/yt/models/asset.rb
```diff
@@ -68,7 +69,7 @@ def metadata_effective
       #   with the asset. You can apply a label to multiple assets to group
       #   them. You can use the labels as search filters to perform bulk updates,
       #   to download reports, or to filter YouTube Analytics.
-      has_attribute :label
+      has_attribute :label, default: []
```
dgb, line 76: Does this change anything? Do we care?

dgb (approved):
This LGTM.

As I mentioned, we're still working on the `as_content_owner` tests, but this looks like any changes there shouldn't affect it too much.

Thanks!

dgb:
Note to self / @kangkyu: this will need a changelog entry.

jcohenho:
Hi @dgb, sorry I was out on vacation but I'm back now. Thanks so much for getting back to me and reviewing this PR.

lib/yt/models/asset.rb
```diff
@@ -106,6 +107,13 @@ def patch_params
           params[:params] = {on_behalf_of_content_owner: @auth.owner_name}
         end
       end
+
+      def get_params
```
kangkyu, line 117: Please add a comment on top, something like
```
# @return [Hash] the parameters to submit to YouTube to get an asset.
# @see https://developers.google.com/youtube/partner/docs/v1/assets/get
```

lib/yt/models/asset.rb
```diff
@@ -6,12 +6,13 @@ module Models
     # Provides methods to interact with YouTube ContentID assets.
     # @see https://developers.google.com/youtube/partner/docs/v1/assets
     class Asset < Base
-      attr_reader :auth
+      attr_reader :auth, :data, :params
```
kangkyu: Do we need to expose `:params` too?

spec/requests/as_content_owner/asset_spec.rb
```diff
@@ -4,13 +4,19 @@
 describe Yt::Asset, :partner do
   describe '.ownership' do
     let(:asset) { Yt::Asset.new id: asset_id, auth: $content_owner }
+    let(:asset_with_metadata_mine) { Yt::Asset.new id: asset_id, auth: $content_owner, params: {fetch_metadata: 'mine'} }
```
kangkyu: the asset has to have metadataMine? how about give it another asset ID or pull out as different `describe`?

lib/yt/models/asset.rb
jcohenho: This won't change anything but I'm just following the same pattern used elsewhere in YT for attributes that return an array of data. Should I remove and let `nil` be the default?

lib/yt/models/asset.rb
jcohenho: No we don't need to. I'll remove these.

spec/requests/as_content_owner/asset_spec.rb
jcohenho: Yes, the asset has to have params in order to fetch data. I can move into its own describe.

lib/yt/models/asset.rb
jcohenho: 👍

spec/requests/as_content_owner/asset_spec.rb
jcohenho: what about a describe titled `'given an asset administered by the content owner that we want to fetch metadata for'`?

jcohenho: I pulled it into a different describe.

lib/yt/models/asset.rb
kangkyu: I agree with @jcohenho

spec/requests/as_content_owner/asset_spec.rb
kangkyu: sorry @jcohenho I wasn't clear. I actually meant by another `describe` it doesn't look great if we put everything under `describe '.ownership'`. And I also thought we should be able to use a separate asset ID for `asset_with_metadata_mine` because not all asset has `metadata_mine` value.

dgb:
Same as your other PR. I changed some stuff with tests so they _should_ run OK on CI. Please try merging in master and hopefully everything goes smoothly.

jcohenho:
Hi @dgb, I added back the default to the label attribute as per @kangkyu's comment. Master is also merged in.

PR #344, opened 2019-03-29
Remove pagination with publishedAfter and publishedBefore

kangkyu:
Currently it loops infinitely when its parent channel has more than
500 videos. (it fetches for 10 pages, and then run the next query
with publishedAfter without page token and - because it doesn't work -
start over from the beginning because publishedAfter is ignored)

YouTube says it's only a temporary issue, but we'll see.
https://issuetracker.google.com/issues/128594402

### DIFF
--- a/lib/yt/collections/videos.rb
+++ b/lib/yt/collections/videos.rb
@@ -8,11 +8,6 @@ module Yt
     # Resources with videos are: {Yt::Models::Channel channels} and
     # {Yt::Models::Account accounts}.
     class Videos < Base
-      def where(requirements = {})
-        @published_before = nil
-        @halt_list = false
-        super
-      end
 
     private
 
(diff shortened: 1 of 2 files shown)


lib/yt/collections/videos.rb
```diff
@@ -93,42 +88,6 @@ def list_params
         end
       end
 
-      def next_page
-        super.tap do |items|
-          halt_list if use_list_endpoint? && items.empty? && @page_token.nil?
-          add_offset_to(items) if !use_list_endpoint? && videos_params[:order] == 'date' && !(videos_params[:for_mine] || videos_params[:for_content_owner])
-        end
-      end
```
mkk-fullscreen, line 101: after removing this we're going to fall back to the definition here: https://github.com/Fullscreen/yt/blob/6e0c51d0f70c03db2f50a6835a1624a30cc473bc/lib/yt/actions/list.rb#L47

kangkyu:
the filters came back https://issuetracker.google.com/issues/128774409 so I'm closing it

PR #345, opened 2019-04-16
Add claims without any parent

kangkyu:


lib/yt/collections/claims.rb
```diff
@@ -58,7 +58,8 @@ def insert_params
       end
 
       def claims_params
-        apply_where_params! on_behalf_of_content_owner: @parent.owner_name
+        owner_name = @parent ? @parent.owner_name : @auth.owner_name
+        apply_where_params! on_behalf_of_content_owner: owner_name
```
kangkyu, line 62: Now I'm not sure if we need this change. Because `owner.claims` does the same thing with `Claims.new(auth: owner)` in this way.

PR #346, opened 2019-04-16
adds match info to new youtube claims API request

efremroberson:
A new requirement to include matching video segments for a claim against an asset is coming down the pipeline for the Youtube partner claims API. 

With this requirement, an API request body _may_ look something like:

```
  {
       'assetId': '<asset_id>',
       'contentType': 'audiovisual',
       'videoId': '<video_id>',
       'policy': {
         'id': '<policy_id>'
       },
       'blockOutsideOwnership': 'true',
       'matchInfo': {
           'matchSegments': [
               {
                   'manual_segment': {
                       'start': '00:00:20.000',
                       'finish': '00:01:20.000'
                   }
               },
               {
                   'manual_segment': {
                       'start': '00:02:00.000',
                       'finish': '00:03:30.000'
                   }
               }
           ]
       }
   }
```

This PR adds `matchInfo` to the body of the "insert claim" request. 

API Documentation: https://developers.google.com/youtube/partner/docs/v1/claims/insert
(Does not yet include information on match info or manual segments.)

Unknown: If each "manual_segment" hash key will be snake cased or will continue with camel case like with other attributes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ For more information about changelogs, check
 
 * [FEATURE] Add `ownership_effective` method to access asset ownership ("effective") via the asset object.
 * [FEATURE] List content owners of others with `content_owner.content_owners`
+* [FEATURE] Add `match_info` to insert claim request.
 
 ## 0.32.3 - 2019-03-15
 
--- a/README.md
+++ b/README.md
@@ -300,6 +300,32 @@ claim.claim_history #=> #<Yt::Models::ClaimHistory ...>
 claim.claim_history.events[0].type #=> "claim_create"
 
 claim.delete #=> true
+
+data = {
+  is_manual_claim: true,
+  content_type: 'audiovisual',
+  asset_id: 'A123123123123123',
+  policy: { id: 'S123123123123123' },
+  video_id: 'myvIdeoIdYT',
+  match_info: {
+    match_segments: [
+      {
+        manual_segment: {
+          start: '00:00:20.000',
+          finish: '00:01:20.000'
+        }
+      },
+      {
+        manual_segment: {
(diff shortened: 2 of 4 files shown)


jcohenho:
Hi @efremroberson, would you be interested in cloning this repo and creating a new gem with me that's better maintained? This repo is no longer maintained and pull requests are not reviewed. Furthermore, a Pull Request which has been created by a non-Fullscreen employee has *never* been accepted by the owners of this repo, and I imagine they never will be.

efremroberson:
Hey @jcohenho, I think this code has just become stale. Fullscreen is only using this gem in a very limited capacity at this point.

I think it would be valuable to create a new gem (perhaps from scratch). I am open to discussing.

efremroberson:
@jcohenho we are revisiting our maintenance policy of this gem now, please sit tight.

kangkyu:
Looks like `camelize_keys!` only camelizes immediate keys so "match_segments" might not be going to become "matchSegments" I think. https://github.com/Fullscreen/yt/blob/master/lib/yt/request.rb#L122

dgb (approved):
👍 

It would be nice to have an integration test for this, but given the state of the partner integration tests (and the fact that this is isn't a "safe" endpoint) let's just do it like this.

kangkyu:
YouTube might be camel-case / snake-case sensitive, that's my concern

dgb:
Yeah, I agree @kangkyu. Let's clarify what the final API is, and make sure it's thoroughly tested.

dgb:
It sounds like the mix of snake and camel case is OK and not without precedent:
https://developers.google.com/youtube/partner/docs/v1/referenceConflicts#matches[]

kangkyu (approved):

PR #348, opened 2019-05-21
Add method to upload reference files via resumable sessions

jcohenho:


dgb:
Hi there,

My name is Derek and I'm now one of the maintainers of the Yt gem. I want to both thank you for your contribution and apologize for the lack of feedback on your pull request. Maintaining an open source project like Yt is a pretty big job, and we're taking steps to prioritize maintenance and development of the project. We're hoping to review community contributions on a regular basis, as well as provide a roadmap for our development plans.

As part of this process we're going through the open PRs and taking action on each one, and for this PR we've chosen to review it. You can expect some more specific feedback in the next week or so.

Thanks again for your contribution, and please let us know if you have any questions or comments.

dgb:
Hey there,

Apologies again for the delay--I think this is something we'd like to merge. I just merged a PR that gets tests working; part of that work included disabling these `as_content_owner` tests because, as you may have noticed, they depend on knowing about and configuring a large number of environment variables, and having a test environment _just so_. If you have any input here on how to make testing easier/better, I'd like to know.

Thanks!

lib/yt/models/content_owner.rb
```diff
@@ -46,6 +46,22 @@ def initialize(options = {})
...
 
+      # Uploads a reference file to YouTube.
+      # @param [String] path_or_url is the video or audio file to upload. Can either be the
+      #   path of a local file or the URL of a remote file.
+      # @param [Hash] params the metadata to add to the uploaded reference.
+      # @option params [String] :asset_id The id of the asset the uploaded reference belongs to.
+      # @option params [String] :content_type The type of content being uploaded.
+      # @return [Yt::Models::Reference] the newly uploaded reference.
+      def upload_reference_file(path_or_url, params = {})
+        file = open path_or_url, 'rb'
```
dgb, line 57: I think this will behave pretty differently if the file comes from a URL vs a path. If it's from a URL, `open` will download the file to a tempfile, whereas if it's a path, we just make a pointer to that file descriptor (and Ruby can ask the OS for its size, etc). Do we care about that?

spec/requests/as_content_owner/content_owner_spec.rb
```diff
@@ -264,6 +265,32 @@
...
+    describe '.upload_reference_file' do
+      let(:asset) { Yt::Asset.new id: ENV['YT_TEST_PARTNER_ASSET_ID'], auth: $content_owner }
+      let(:match_policy) { Yt::MatchPolicy.new asset_id: ENV['YT_TEST_PARTNER_ASSET_ID'], auth: $content_owner }
+
+      let(:upload_params) { {asset_id: asset.id, content_type: 'video'} }
+      let(:reference) { $content_owner.upload_reference_file path_or_url, upload_params }
+      after { reference.delete }
+
+      before do
+        asset.ownership.update(assetId: asset.id) && asset.ownership.obtain!
```
dgb, line 278: I'm thinking of moving some of this expensive/complicated "fixture" creation to outside the tests themselves.

spec/requests/as_content_owner/content_owner_spec.rb
```diff
@@ -264,6 +265,32 @@
...
+        match_policy.update policy_id: ENV['YT_TEST_PARTNER_POLICY_ID']
+      end
+
+      context 'given the path to a local video file' do
+        let(:path_or_url) { File.expand_path '../reference_video.mp4', __FILE__ }
```
dgb: This is a pretty big file (23.4MB). Is there something else we can do?

jcohenho: A video needs to be at a minimum 20 seconds long in order to create a reference file, maybe I can compress and upload a lower quality version.

spec/requests/as_content_owner/content_owner_spec.rb
jcohenho: Should we do this in a separate PR or in this one?

lib/yt/models/content_owner.rb
jcohenho: So the main difference is that with a URL, we have a tempfile left over in the directory? It doesn't matter for my use case, as tempfiles get cleaned up automatically in production. What do you think? This is the same exact code taken from the [upload_video](https://github.com/Fullscreen/yt/blob/master/lib/yt/models/account.rb#L71) method in `account.rb`.

spec/requests/as_content_owner/content_owner_spec.rb
dgb: Oh, this is something I'm planning to do in the future. Just wanted to give you a heads up.

spec/requests/as_content_owner/content_owner_spec.rb
dgb: Hrm, that's too bad. I wonder if we need to have an explicit test here for the local file case if we're just using `open-uri`...in a way we'd just be testing that open-uri works.

jcohenho: Sure, I can change the test to use a remote URL instead.

jcohenho: @dgb Forgot that I already had the remote URL test case here. Updated to remove local test case and file.

PR #349, opened 2019-06-07
Use Ruby 2.6 on Travis

dgb:


dgb:
We can discuss what versions we want to support later; this is just to get tests running at all on Travis.

kangkyu (approved):

PR #350, opened 2019-06-13
Update tests

dgb:
This pull requests gets tests in a shape where they're all passing (at least on my machine 😄 ).

In all seriousness, it mostly updates some ids to correspond to resources on YouTube that exist (since some channels, playlists, etc went away). It also marks a few tricky tests (and _all_ `as_content_owner` tests) to be skipped--we need to rethink how some of these tests work, so I'll create follow on tickets for this work.

It also ads _some_ documentation for the integration test env variables. We should probably clean up the setup here; I have some thoughts about this that we can discuss offline.


spec/requests/as_account/playlist_spec.rb
```diff
@@ -121,9 +121,9 @@
 
       specify 'updates them replacing angle brackets with similar unicode characters accepted by YouTube' do
         expect(update).to be true
-        expect(playlist.title).to eq 'Yt Test ‹ ›'
-        expect(playlist.description).to eq '‹ ›'
-        expect(playlist.tags).to eq ['‹tag›']
+        expect(@my_playlist.title).to eq 'Yt Test ‹ ›'
+        expect(@my_playlist.description).to eq '‹ ›'
+        expect(@my_playlist.tags).to eq ['‹tag›']
```
kangkyu: Is this change necessary? Existing playlist (`playlist` with `id`) should be enough to fetch info I think

spec/requests/as_account/video_spec.rb
```diff
@@ -396,7 +396,7 @@
   # @see https://developers.google.com/youtube/v3/docs/videos#fileDetails
   # @see https://developers.google.com/youtube/v3/docs/videos#processingDetails.fileDetailsAvailability
   context 'given one of my own videos' do
-    let(:id) { 'nHz3FnAMH3U' }
+    let(:id) { $account.videos.first.id }
```
kangkyu, line 399: To match the test code, we better say "given one of my own videos, not *available*" like this. Or find an *available* video and put id of that video

spec/requests/as_account/playlist_spec.rb
dgb: It wasn't working with `playlist`, and all the other tests in this section use `@my_playlist`. I didn't investigate much further than that.

spec/requests/as_account/video_spec.rb
dgb: Sure. Ideally we'd have some sort of fixture script and we could look up the video by title or something.

spec/requests/as_account/playlist_spec.rb
kangkyu: but it is testing if the playlist is updated or not, right? If it wasn't working then... I'm not sure, when I tried before this test was working

dgb: Sure, but `update` runs on `@my_playlist`, not `playlist`. They're two different objects, so the updates won't be reflected on `playlist`.

kangkyu:
I think it would be nice if we squash the commits before merge!

kangkyu (approved):

PR #354, opened 2019-06-24
Add Resource.new(url: <url>) syntax back to yt

kangkyu:
I moved much code from url.rb to resource.rb to keep `url.resource` still work

### DIFF
--- a/lib/yt/models/resource.rb
+++ b/lib/yt/models/resource.rb
@@ -12,7 +12,9 @@ module Yt
 
       # @!attribute [r] id
       #   @return [String] the ID that YouTube uses to identify each resource.
-      attr_reader :id
+      def id
+        @id ||= @match['id'] || fetch_channel_id
+      end
 
     ### STATUS ###
 
@@ -42,7 +44,12 @@ module Yt
 
       # @private
       def initialize(options = {})
-        @id = options[:id]
+        if options[:url]
+          @url = options[:url]
+          @match = find_pattern_match
+        else
+          @id = options[:id]
+        end
         @auth = options[:auth]
         @snippet = Snippet.new(data: options[:snippet]) if options[:snippet]
         @status = Status.new(data: options[:status]) if options[:status]
@@ -50,7 +57,11 @@ module Yt
 
       # @private
       def kind
-        self.class.to_s.demodulize.underscore
+        if @url
+          @match[:kind].to_s
+        else
+          self.class.to_s.demodulize.underscore
+        end
       end
 
       # @private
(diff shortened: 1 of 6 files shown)


lib/yt/models/resource.rb
```diff
@@ -12,7 +12,13 @@ class Resource < Base
 
       # @!attribute [r] id
       #   @return [String] the ID that YouTube uses to identify each resource.
-      attr_reader :id
+      def id
+        if @url
```
dgb: Do we need this if/else here? In your code if you don't instantiate with a url, you've got `@id` set already.

spec/models/resource_spec.rb
```diff
@@ -4,6 +4,13 @@
 describe Yt::Resource do
   subject(:resource) { Yt::Resource.new attrs }
 
+  context 'given a resource initialized with a URL (containing an ID)' do
+    let(:attrs) { {url: 'youtu.be/9bZkp7q19f0'} }
+
+    it { expect(resource.id).to eq '9bZkp7q19f0' }
```
dgb, line 10: Won't this make an HTTP request? If so, it seems like this test belongs in the `requests` dir.

lib/yt/models/resource.rb
```diff
@@ -42,15 +48,20 @@ def unlisted?
...
+        end
         @auth = options[:auth]
         @snippet = Snippet.new(data: options[:snippet]) if options[:snippet]
         @status = Status.new(data: options[:status]) if options[:status]
       end
 
       # @private
       def kind
-        self.class.to_s.demodulize.underscore
+        (@match[:kind].to_s if @url) || self.class.to_s.demodulize.underscore
```
dgb: Can we change this to be an `if/else`? It's kind of hard to read as-is.

```ruby
if @url 
  @match[:kind].to_s
else
  self.class.to_s.demodulize.underscode
end

File: lib/yt/models/resource.rb [hunk 4]
```diff
@@ -66,8 +77,63 @@ def update(attributes = {})
...
+      # @return [Array<Regexp>] patterns matching URLs of YouTube channels.
+      CHANNEL_PATTERNS = [
+        %r{^(?:https?://)?(?:www\.)?youtube\.com/channel/(?<id>UC[a-zA-Z0-9_-]{22})},
+        %r{^(?:https?://)?(?:www\.)?youtube\.com/(?<format>c/|user/)?(?<name>[a-zA-Z0-9_-]+)}
+      ]
+
     private
 
+      def find_pattern_match
+        patterns.find(-> {{kind: :unknown}}) do |kind, regex|
```
dgb: I know this is stuff you moved around, but I find this kind of hard to follow.

spec/models/resource_spec.rb
kangkyu: Does it? I think it doesn't because it's a video id

dgb: I might have just misunderstood how the url part works :)

kangkyu: By the way, we might need add more tests in `requests/` directory, for this change

PR #355, opened 2019-06-26
Fix claim match segments

dgb:
The `matchSegments` sub-key should be camelCase, not snake_case per the
documentation.


kangkyu (approved):

kangkyu:
following https://github.com/Fullscreen/yt/pull/346

PR #374, opened 2020-01-14
[WIP] Bug Fix: no videos should be an empty array always

kangkyu:


kangkyu:
we might better raise `NoItemsError` when there's nothing returned. I'm not sure.

kangkyu:
Closing because we can expect `@where_params` becomes empty in the next call. (See https://github.com/Fullscreen/yt/pull/45)

PR #375, opened 2020-02-06
Delete playlist items via content owner

dgb:


lib/yt/models/playlist_item.rb
```diff
@@ -101,6 +101,12 @@ def resource_id
         {kind: 'youtube#video', videoId: video_id}
       end
 
+      def delete_params
+        super.tap do |params|
+          params[:params].merge! @auth.playlist_items_params
```
kangkyu, line 107: I think method name `playlist_items_params` can be different? Actually it is used in collections file
https://github.com/Fullscreen/yt/blob/master/lib/yt/collections/playlist_items.rb#L44

How about `playlist_item_params`

dgb: We chatted about this offline, but for the viewers at home: `delete_params` is invoked by the `Delete` action when you call `do_delete`, and it has a default implementation (hence the `super`).

kangkyu (approved):

PR #376, opened 2020-02-06
:vhs: Record the test suite with VCR

dgb:
Changes the test setup to use VCR on any tests that hit the YouTube API.
Also simplifies the use of authentication credentials, and more
granularly disables some tests. Finally, changes `Request` and `Reports`
to make their retry times stubbable in order to speed up test runs.


spec/errors/request_error_spec.rb
```diff
@@ -1,5 +1,5 @@
 require 'spec_helper'
-require 'yt/errors/request_error'
+require 'yt'
```
kangkyu: why do we need this change?

.rspec
```diff
@@ -1,6 +1,3 @@
---format documentation
```
kangkyu, line 1: I think it's nice to have documentation format, but it's your call

spec/requests/as_account/playlist_spec.rb
```diff
@@ -206,7 +206,11 @@
   end
 
   context 'given one of my own playlists that I want to get reports for' do
-    let(:id) { $account.channel.playlists.first.id }
+    let(:id) { test_account.channel.playlists.first.id }
+
+    before do
+      allow(Date).to receive(:today).and_return(Date.new(2020, 2, 5))
```
kangkyu, line 195: why do we need this?

yt.gemspec
```diff
@@ -1,4 +1,3 @@
-# coding: utf-8
```
kangkyu, line 1: what is the difference?

spec/errors/request_error_spec.rb
dgb: I think this was a change for something else I was doing. Doesn't look necessary now. Reverted in ea919e9

kangkyu (approved):

.rspec
dgb: I don't really like it for normal test runs since it makes it impossible to see where the progress is. You can always pass it if you want to see the doc format too.

spec/requests/as_account/playlist_spec.rb
dgb: Because this test is exercising a function that (ultimately) calls `Date.today`. We need to fix that so that it's deterministic for VCR (otherwise it'll see a request for a different day and fail).

yt.gemspec
dgb: It's necessary for old Ruby (< 2.0 IIRC), which we should drop support for.

kangkyu: thank you!

PR #377, opened 2020-02-06
Update video via content owner

kangkyu:


lib/yt/models/video.rb
```diff
@@ -667,6 +667,14 @@ def update_parts
         {snippet: snippet, status: {keys: status_keys}}
       end
 
+      # For updating video with content owner auth.
+      # @see https://developers.google.com/youtube/v3/docs/videos/update
+      def update_params
+        super.tap do |params|
+          params[:params] = {on_behalf_of_content_owner: auth.owner_name} if auth.owner_name
```
dgb: What do you think about extending the pattern we have and adding `video_params` to `ContentOwner` and `Account`. That way this doesn't need to know or care what auth is, and can just look something like:

```ruby
def update_params
  super.tap do |params|
    params[:params].merge!(auth.video_params)
  end
end
```

kangkyu: Unrelated but `params[:params]` can be nil so we can `params[:params] ||= {}`

dgb (approved):

PR #378, opened 2020-02-07
upload thumbnail, update playlist, and insert playlist item via content owner

kangkyu:


lib/yt/models/video.rb
```diff
@@ -643,8 +643,11 @@ def upload_path
       # @private
       # Tells `has_many :resumable_sessions` what params are set for the object
       # associated to the uploaded file.
+      # https://developers.google.com/youtube/v3/docs/thumbnails/set
       def upload_params
-        {video_id: id}
```
dgb, line 647: This indicates to me that we should probably rethink the pattern we're extending and the hierarchy here. Many of these are calling `super` to get the params on the corresponding "action". Maybe this is just an exceptional case though...

dgb (approved):
LGTM

We can add some tests after the VCR stuff is merged for what's in here:
* inserting playlist items
* updating playlist items
* uploading thumbnails

PR #384, opened 2020-04-10
Limit retries on refresh tokens

dgb:
Sometimes YouTube gives a response we're not expecting when refreshing
an access token. We shouldn't keep attempting to refresh if this
happens.

### DIFF
--- a/lib/yt/request.rb
+++ b/lib/yt/request.rb
@@ -196,7 +196,7 @@ module Yt
     # - when the user has reached the quota for requests/second, and waiting
     #   for a couple of seconds might solve the connection issues.
     def run_again?
-      refresh_token_and_retry? ||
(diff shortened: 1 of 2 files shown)


kangkyu (approved):

lib/yt/request.rb
```diff
@@ -196,7 +196,7 @@ def parse_response!
     # - when the user has reached the quota for requests/second, and waiting
     #   for a couple of seconds might solve the connection issues.
     def run_again?
-      refresh_token_and_retry? ||
+      refresh_token_and_retry? && sleep_and_retry?(3) ||
```
dgb, line 199: I'd like to make the retry calls iterative, but that's an improvement we can make later.

PR #390, opened 2020-09-08
Remove comment_count metric of a channel, by YouTube change

kangkyu:
I received an email from no-reply@youtube.com on Aug 12:

Screen Shot 2020-08-17 at 11 13 28 AM

I think this is the only place to be removed accordingly as far as Yt gem is concerned.


mkk-fullscreen:
looks like line 21 in the readme should go too

dgb (approved):
Let's cleanup the Readme, but otherwise LGTM

PR #397, opened 2020-11-17
Allow nil id and require URL

dgb:
Make sure that instantiating models with a `nil` ID is defined. Also
require the URL model in Yt.


lib/yt/models/resource.rb
```diff
@@ -13,7 +13,11 @@ class Resource < Base
       # @!attribute [r] id
       #   @return [String] the ID that YouTube uses to identify each resource.
       def id
-        @id ||= @match['id'] || fetch_channel_id
+        if @id.nil? && @match && @match[:kind] == :channel
```
dgb, line 16: I would prefer to put this in `Channel`, but the way `URL` works, it instantiates a base `Resource` here, so fixing that is a much bigger refactor.

PR #401, opened 2021-01-15
Don't retry on quota errors

dgb:
They say that insanity is doing the same thing over and over and
expecting different results...

This patch removes retry behavior on quota issues because:

* this isn't useful when daily quota has been reached
* in the case of 100s quota, you'd need to be attempting requests ~61s
into the quota cool-off in order for the final retry to succeed. That
seems pretty unlikely.

So for all practical purposes, when quota is reached, the library stalls
for the maximum timeout time (39s by my math). This is further
problematic when making requests in environments like Heroku that have
a 30s timeout for web request-response cycles.

Let's do the sane thing and just raise an error when quota is reached.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
+## 0.33.4 - 2021-01-15
+
+* [REMOVAL] remove retry for quota errors
+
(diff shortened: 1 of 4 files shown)


prajaktaedcast:
hi @dgb,

We are using YT gem and it is pulling nonrelevant videos which are not included in channels while pulling videos from channel by passing channel_id. We only getting exceed quota in our app logger. can you explain because of Quota exceeds error we are getting nonrelevant data???

also will this fix "remove retry.." will help us??

PR #403, opened 2021-03-05
Enable bundler caching for travis

Taher-Ghaleb:
Would be interested to know why Travis caching hasn't been enabled in this repository. Thank you.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: ruby
 notifications:
   email: true
+cache: bundler
 matrix:
   include:
   - rvm: 2.6.3


kangkyu:
I'm not sure if we still use Travis. Do we? @justinhoward I guess we might be able to remove the `.travis.yml` file if not

justinhoward:
@kangkyu We could consider switching to GH actions. From what I understand, Travis CI for Open Source is defunct.

kangkyu:
Thank you @justinhoward closing because we don't consider continue to use travis

PR #408, opened 2021-12-06
Add region_restricted? for video

dlackty:
I followed Video#age_restricted? and implement this helper.

Ref: https://developers.google.com/youtube/v3/docs/videos#properties

### DIFF
--- a/lib/yt/models/video.rb
+++ b/lib/yt/models/video.rb
@@ -271,6 +271,11 @@ module Yt
         content_detail.licensed_content || false
       end
 
+      # @return [Boolean] whether the video is only viewable in limited countries.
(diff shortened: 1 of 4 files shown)


lib/yt/models/video.rb
```diff
@@ -271,6 +271,11 @@ def licensed?
         content_detail.licensed_content || false
       end
 
+      # @return [Boolean] whether the video is only viewable in limited countries.
+      def region_restricted?
+        content_detail.region_restriction
+      end
```
kangkyu, line 277: what does it return when not region restricted? In my opinion we'd better return `true` when region restricted, `false` when not.

PR #412, opened 2022-05-08
Removed dependency on the deprecated open-uri in Ruby 3

kalemi19:
Relying on URI.open() instead

https://stackoverflow.com/a/65939036


declantraynor:
Can this fix be released? My team have recently upgraded to Ruby 3 and our upload features are broken for external URLs. Thanks @kalemi19 for the patch!

kangkyu:
Thank you!

PR #419, opened 2023-10-03
Captions associated with a video

### LINKED ISSUE #413: Captions API
Using this gem, is there a way to download a video's [captions](https://developers.google.com/youtube/v3/docs/captions)? If not, would be awesome if you could add that feature. Thanks!

kangkyu:
After this change, we can do something like this 

```rb
v = Yt::Video.new(id: "8ompJkDKEEI")
c = v.captions.first
c.language
# => "es"
```

closes #413


justinhoward:
The captions addition looks good to me. Does the removal of cassettes mean that tests will now only run whenever someone has set the following env variables? And they will always test against live data?

```
YT_TEST_CLIENT_ID
YT_TEST_CLIENT_SECRET
YT_TEST_API_KEY
YT_TEST_REFRESH_TOKEN
```

Did you test running the suite from scratch to see if that works?

kangkyu:
Yes, I ran test with those env variables and it passed. As you may know I couldn't run `as_content_owner` test because I don't have access to any content owner on YouTube. By the way, that test actually currently disabled like https://github.com/nullscreen/yt/blob/master/.rspec#L2 so I was able to simply run `rspec`

```sh
yt (captions) $ rspec
Run options: exclude {:slow=>true}

Randomized with seed 53975
....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Finished in 1 minute 29.11 seconds (files took 0.75296 seconds to load)
724 examples, 0 failures

Randomized with seed 53975

Coverage report generated for RSpec to /Users/kangkyu/code/yt/coverage. 5678 / 6039 LOC (94.02%) covered.
```

justinhoward (approved):
Nice!

PR #422, opened 2023-10-04
Use new oauth2 endpoints

kangkyu:


kangkyu:
I hope to run all the tests fresh and pass before we will be able to merge this, because these endpoints are about auth and need to make sure all parts of this gem work. But do we really need that? I don't think there's much chance to run as_content_owner test because currently I don't have access to any account with content owners. On the other hand I believe this url change will work and we are generally not always 100% sure.

Let's just merge if you are ok with that @justinhoward

justinhoward:
@kangkyu That sounds like a reasonable plan to me. We may simply want to unofficially drop support for as_content_owner until someone with that access offers support.

justinhoward (approved):

kangkyu:
I am including my updated test codes here. Thanks!

PR #426, opened 2023-12-23
Release a new version

kangkyu:
For someone has ownership

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,18 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
-## 0.34.0 - 2023-12-22
+## 0.34.0 - 2024-08-25
 
-* Ruby 2.1 required
-* [IMPROVEMENT] Use new oauth2 endpoints to be secure
-* [REMOVAL] Remove `Playlist#tags` as 'snippet.tags[]' deprecated
-* [FEATURE] Add captions associated with a video
-* [BUGFIX] Remove dependency on the deprecated open-uri in Ruby 3 (thank you @kalemi19)
+* Ruby 2.1+ required https://github.com/nullscreen/yt/pull/425
+* [IMPROVEMENT] Use new oauth2 endpoints https://github.com/nullscreen/yt/pull/422
+* [REMOVAL] Remove `Playlist#tags` as 'snippet.tags[]' deprecated https://github.com/nullscreen/yt/pull/421
+* [FEATURE] Add captions associated with a video https://github.com/nullscreen/yt/pull/419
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
+## 0.34.0 - 2023-12-22
+
+* Ruby 2.1 required
```
shirts: Ruby 2.1 is not required. 2.1+ is

```suggestion
* Ruby 2.1+ required
```

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ For more information about changelogs, check
...
+* [IMPROVEMENT] Use new oauth2 endpoints to be secure
```
shirts: IMO drop the `to be secure`

Can we link to that here?

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ For more information about changelogs, check
 [Keep a Changelog](http://keepachangelog.com) and
 [Vandamme](http://tech-angels.github.io/vandamme).
 
+## 0.34.0 - 2023-12-22
+
```
shirts, line 10: I personally think that links should be added for each item that's changed. It'll make life easier for the people that read this

lib/yt/version.rb
```diff
@@ -1,3 +1,3 @@
 module Yt
-  VERSION = '0.33.4'
+  VERSION = '0.34.0'
```
shirts, line 2: [The readme says](https://github.com/nullscreen/yt/blob/master/README.md#how-to-install) this gem uses semver. This could be a breaking change for anyone on Ruby < 2.1. Although the gem isn't even at 1.0, typically there's leniency with breaking changes. 

How do we feel about this? The readme is misleading at this point IMO

@justinhoward thoughts?

shirts (changes requested):

kangkyu:
Thank you for your review @shirts I'll get back to your comments

lib/yt/version.rb
kangkyu: Actually this gem has been increasing the second number when there's a breaking change, just the patch number increase for non-breaking changes, and never increased first number of version 😅 If you want a differently numbered version value, we can do that from now on of course.

shirts: I get that but it's just internal knowledge of how the version bumping works instead of truly following semver which the readme describes and people may depend on.

Should we add a blurb about that in the readme?

kangkyu: I wish it is ok to update CHANGELOG better, to clarify that this is a breaking upgrade.

shirts (approved):

PR #428, opened 2024-01-23
Prepare CI with GitHub actions for publishing

justinhoward:
Testing to see what needs to be in place to prepare for automated publishing to rubygems. I assume we'll want tests to pass, so some changes or filtering may need to be done to the test suite.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,76 @@
...
+    branches: [master]
+  pull_request:
+    branches: ['**']
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby:
```
shirts: Should 3.3 be included?

justinhoward: Good idea.

kangkyu:
@justinhoward If not for checking tests passing, and for solely gem release, why don't you remove the test steps from the action?

justinhoward:
Disabled tests in CI for now. I'm not sure how to solve the testing problem, but perhaps its best to move forward with releases anyway.

yt.gemspec
```diff
@@ -5,32 +5,25 @@ require 'yt/version'
 Gem::Specification.new do |spec|
   spec.name          = "yt"
   spec.version       = Yt::VERSION
-  spec.authors       = ["Claudio Baccigalupo"]
-  spec.email         = ["claudio@fullscreen.net"]
+  spec.authors       = ["Nullscreen"]
+  spec.email         = ["nullscreen.code@gmail.com"]
```
kangkyu, line 9: "nullscreen" has rubygems account?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,82 @@
...
+      - run: bundle exec yardoc --fail-on-warning
+
+  check_version:
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '2.7'
```
kangkyu, line 61: Does this mean the check happens for ruby of version `2.7`? That version EOL was like over a year ago

yt.gemspec
kangkyu: I see it here, sorry https://rubygems.org/profiles/Nullscreen

.github/workflows/ci.yml
justinhoward: This gem supports down to 2.1 at the moment according to the gemspec. Is using 2.7 to run the yard check a problem?

kangkyu: I see.

kangkyu (approved):

PR #430, opened 2024-08-30
Add access to an account's sub ID in the v3 API

jkowens:
The v3 API no longer appears to return an "id" value, but instead a "sub" value that is the account's unique identifier.

See: https://developers.google.com/identity/openid-connect/openid-connect

Should the id method be deleted? @kangkyu what are your thoughts?

### DIFF
--- a/lib/yt/models/account.rb
+++ b/lib/yt/models/account.rb
@@ -14,6 +14,10 @@ module Yt
       #   @return [String] the (Google+) account’s ID.
       delegate :id, to: :user_info
 
+      # @!attribute [r] sub
+      #   @return [String] the (Google) account’s unique ID.
+      delegate :sub, to: :user_info
+
       # @!attribute [r] email
       #   @return [String] the account’s email address.
       delegate :email, to: :user_info
--- a/lib/yt/models/user_info.rb
+++ b/lib/yt/models/user_info.rb
@@ -11,6 +11,7 @@ module Yt
       end
 
       has_attribute :id, default: ''
+      has_attribute :sub, default: ''
       has_attribute :email, default: ''
       has_attribute :verified_email, default: false, camelize: false
       has_attribute :name, default: ''
--- a/spec/models/account_spec.rb
+++ b/spec/models/account_spec.rb
@@ -16,6 +16,18 @@ describe Yt::Video do
     end
   end
 
+  describe '#sub' do
+    context 'given a user info with an ID' do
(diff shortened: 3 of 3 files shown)


kangkyu:
For what request do you get "sub"? Let me check later today

jkowens:
It looks like `sub` is returned instead of `id` when fetching the account user info.

kangkyu:
I see @jkowens `sub` property is in response JSON of `'www.googleapis.com/oauth2/v3/userinfo'` request. (I think this endpoint is not documented on their API documentation, so any change happened might have not been announced)

To resolve, I think we should add new field and deprecate old field for this gem's users. So we remove `id` in the future, but add `sub` right now. We don't have to use alias to keep the name of old field IMO and that `id` field should be removed if not returned at all.

```json
{
  "sub": "123456789012345678901",
  "name": "Jimmy Example",
  "given_name": "Jimmy",
  "family_name": "Example",
  "picture": "https://lh3.googleusercontent.com/a/image-url",
  "email": "jimmy.example@gmail.com",
  "email_verified": true
}
```

jkowens:
@kangkyu I updated the PR to deprecate the `id` methods. I added an `#id` method on UserInfo that calls sub, but also prints a deprecation warning. I thought that would give people a chance to update their code before the `id` method is deleted. Is that ok?

kangkyu:
I think I still need to check this line 
https://github.com/nullscreen/yt/blob/master/spec/requests/as_content_owner/account_spec.rb#L6
It is totally unrelated, and we don't need the `id` key when initialize `Yt::Account` at all I think, but I want to make sure. Give me some moment

kangkyu:
Well, `ContentOwner` inherits `Account`.
https://github.com/nullscreen/yt/blob/master/lib/yt/models/content_owner.rb#L8

It will take time to look into YouTube Content Owner API part of this gem.. We might be using `id` in some `ContentOwner` object, or not, I do not know.

So, how about this @jkowens, for this PR:

😅 We can leave `id` field just as it is. It will be ok because it is not an unexpected behavior for all of us. I mean there's no harm to have `id` always stay as an empty string. If we still can have the value of `sub`, then we are good I think. And then we will be able to remove `id` or anything we can do at some point in the future.

How do you think? Let me know if you agree.

jkowens:
@kangkyu that's good with me. I updated the PR to simply add access to the `sub` attribute. I'll let you maintainers worry about the rest 😄

kangkyu:
Thank you!

PR #433, opened 2024-10-09
Update account.rb

lupoweb:
resolve private method `open' called for URI:Module


kangkyu:
If we make this change, the code will not work with Ruby 3 anymore. FYI

kangkyu:
Sorry just sanity check, do you use Ruby 3? @lupoweb

kangkyu:
superseded by https://github.com/nullscreen/yt/pull/438

PR #447, opened 2025-06-27
Get channel id from channel url

kangkyu:
- Use "prompt=consent" and not use "approval_prompt=force"
- Download caption of video

### DIFF
--- a/lib/yt/actions/get.rb
+++ b/lib/yt/actions/get.rb
@@ -14,7 +14,7 @@ module Yt
     private
 
       def get_request(params = {})
-        @list_request = Yt::Request.new(params).tap do |request|
+        @get_request = Yt::Request.new(params).tap do |request|
           print "#{request.as_curl}\n" if Yt.configuration.developing?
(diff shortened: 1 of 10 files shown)


kangkyu:
@justinhoward Would you review this pr

lib/yt/actions/get.rb
```diff
@@ -14,7 +14,7 @@ def get
     private
 
       def get_request(params = {})
-        @list_request = Yt::Request.new(params).tap do |request|
+        @get_request = Yt::Request.new(params).tap do |request|
```
justinhoward, line 17: Nit: This assignment serves no purpose right?

justinhoward (approved):
Nice! Looks good to me

lib/yt/actions/get.rb
kangkyu: Yes you are right, the `@get_request` is not used anywhere. We need to raise an error using it in the message, when request failed.

PR #438, opened 2025-03-05
Use open-uri to open uri as file

kangkyu:
Fixes #432

PR #448, opened 2025-08-23
Fix upload_video "private method 'open' called for an instance of URI::Generic" error

mrj:
Calling `Account#upload_video` on `master` with a file path string (no `file://` URI scheme), I get a `private method 'open' called for an instance of URI::Generic` error that wasn't fixed by #438.

This PR [changes](/nullscreen/yt/blob/16944ffb0876389a07c86a72d3e7a06f1455bba6/lib/yt/models/account.rb#L77)
```
file = URI.parse(path_or_url).open
```
to
```
file = URI.open(path_or_url)
````
[`open`](/ruby/ruby/blob/5b5b5b3af555a0349c3d7e4e2a50f36ecd848560/lib/open-uri.rb#L23) automatically parses first when its argument is a string-with-scheme, and defers to `Kernel#open` for paths.

I'm testing on Ruby 3.4.5. But looking at [the history of `open-uri.rb`](/ruby/ruby/commits/master/lib/open-uri.rb), this call to `open` should work on even ancient Ruby versions.

### DIFF
--- a/lib/yt/models/account.rb
+++ b/lib/yt/models/account.rb
@@ -74,7 +74,7 @@ module Yt
       # @option params [Boolean] :self_declared_made_for_kids The video’s made for kids self-declaration.
       # @return [Yt::Models::Video] the newly uploaded video.
       def upload_video(path_or_url, params = {})
-        file = URI.parse(path_or_url).open
+        file = URI.open(path_or_url)
         session = resumable_sessions.insert file.size, upload_body(params)
 
         session.update(body: file) do |data|


kangkyu:
According to your comment above, we need basically revert https://github.com/nullscreen/yt/commit/746a1b335b08151c4d9720c8af4924019159bf37 

I don't exactly remember but "open-uri" has been changed a bit. Your change works for any version of Ruby? @mrj

mrj:
The `require 'open-uri'` in  746a1b3 is definitely needed. But I wasn't aware that it also made the two changes from `open` to `parse + open`. These additional changes do break path URLs for me. I'll ask the committer about their motivation for those changes.

kangkyu (approved):

kangkyu:
Thank you @mrj let me know if there's anything you want to add or change. I am going to merge if you are ok

mrj:
Thanks Kang-Kyu. This PR only fixes `upload_video` not `upload_thumbnall`. I'll add that first.

mrj:
Thanks for committing this. In #449 I've done the same with `upload_thumbnail`.

PR #458, opened 2026-05-07
Resumable upload of video

kangkyu:
* FIX: When refresh, expired_at becomes Time (this was an existing bug in the code)


claudiob (approved):
So much code 😬

kangkyu:
@claudiob yes it's too much 😊

I wanted token refresh in the middle of the "stream" so used net/http directly, I think that's why

PR #459, opened 2026-06-11
Handle missing duration for video premieres

marcoroth:
The YouTube API omits `contentDetails.duration` while a video is an ongoing premiere. The default value `0` was then passed to `to_seconds`, which expects an ISO 8601 string, raising  `undefined method 'match' for an instance of Integer`:

```
duration error: NoMethodError: undefined method 'match' for an instance of Integer
  /Users/marcoroth/.local/share/mise/installs/ruby/4.0.1/lib/ruby/gems/4.0.0/gems/yt-0.34.1/lib/yt/models/content_detail.rb:62:in 'Yt::Models::ContentDetail#to_seconds'
  /Users/marcoroth/.local/share/mise/installs/ruby/4.0.1/lib/ruby/gems/4.0.0/gems/yt-0.34.1/lib/yt/models/content_detail.rb:18:in 'block in <class:ContentDetail>'
```

This pull request fixes this so that `Video#duration` now returns `0` and `Video#length` `"00:00:00"` in that case.

### DIFF
--- a/lib/yt/models/content_detail.rb
+++ b/lib/yt/models/content_detail.rb
@@ -14,6 +14,8 @@ module Yt
       end
 
       # @return [Integer] the duration of the video (in seconds).
+      #   Returns 0 for ongoing live broadcasts and premieres, since the
+      #   YouTube API does not report a duration until they have ended.
       has_attribute :duration, default: 0 do |value|
         to_seconds value
       end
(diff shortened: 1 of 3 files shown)


kangkyu:
Curious -- what if we don't call `duration` for those premieres? @marcoroth

lib/yt/models/content_detail.rb
```diff
@@ -14,6 +14,8 @@ def initialize(options = {})
       end
 
       # @return [Integer] the duration of the video (in seconds).
+      #   Returns 0 for ongoing live broadcasts and premieres, since the
+      #   YouTube API does not report a duration until they have ended.
       has_attribute :duration, default: 0 do |value|
         to_seconds value
       end
```
kangkyu, line 21:
```rb
has_attribute :duration, default: "PT0H0M0S"
```
Possibly, this makes more sense? @claudiob @marcoroth

claudiob: Whatever you prefer 😁

kangkyu:
I have a sample code here #463

kangkyu:
Closing superseded by #463

PR #462, opened 2026-07-18
Support for Live Broadcasts

### LINKED ISSUE #461: Support for LiveBroadcasts
Hi,

I'm new to Yt so I could be wrong, I could not find a way to schedule live broadcasts and define their settings.

My use case is to schedule several live events and attach them to a playlist.

I have a working implementation using [https://developers.google.com/youtube/v3/live/docs/liveBroadcasts/insert#request](https://developers.google.com/youtube/v3/live/docs/liveBroadcasts/insert#request)

gmarziou:
This PR implements LiveBroadcast resource and collection.

To create a alive broadcast:

```ruby
    params = {
      title: title, description: description, privacy_status: privacy_status,
      scheduled_start_time: scheduled_start_time, scheduled_end_time: scheduled_end_time,
      default_language: "en"
    }
    account.create_live_broadcast(**params)
 ```

To list all live broadcasts of an account/channel:

```ruby
account(email).live_broadcasts
```

Fix #461


kangkyu:
@gmarziou thank you for opening a pr! I will try and get back to you

lib/yt/models/account.rb
```diff
@@ -121,11 +121,32 @@ def resumable_upload_video(path_or_url, params = {})
       # @option params [String] :privacy_status The new playlist’s privacy
       #   status. Must be one of: private, unscheduled, public.
       # @example Create a playlist titled "My favorites".
-      #   account.create_playlist title: 'My favorites'
+      #   account.*create_playlist title: 'My favorites'
```
kangkyu: This comment update no need

lib/yt/models/account.rb
```diff
@@ -174,11 +195,15 @@ def create_playlist(params = {})
...
+      # @!attribute [r] live_broadcasts
+      #   @return [Yt::Collections::LiveBroadcasts] the live broadcasts owned by the account.
+      delegate :live_broadcasts, to: :channel
+
       # @!attribute [r] videos
       #   @return [Yt::Collections::Videos] the videos owned by the account.
       has_many :videos
 
-      # @!attribute [r] subscribers
+    # @!attribute [r] subscribers
```
kangkyu: This comment update no need

lib/yt/collections/live_broadcasts.rb
```diff
@@ -0,0 +1,68 @@
...
+          attributes[:status] = data['status']
+          attributes[:content_details] = data['contentDetails']
+        end
+      end
+
+      def insert_parts
+        snippet = {keys: [:title, :description], sanitize_brackets: true}
+        status = {keys: [:privacy_status]}
+        {snippet: snippet, status: status}
+      end
```
kangkyu, line 41: I think you might have wanted to use `insert_parts` in `insert` method in this class.

lib/yt/models/live_broadcast.rb
```diff
@@ -0,0 +1,88 @@
...
+      #   @return [Time] the date and time that the broadcast was published.
+      delegate :published_at, to: :snippet
+
+      # @return [String] the ID that YouTube assigns and uses to uniquely
+      #   identify the live_broadcast.
+      has_attribute :id
+
+      def insert
+        do_insert
+      end
```
kangkyu: I think this `insert` method may not be used.

kangkyu (commented):
`params` and `body` can be handled differently, I think we can use more existing methods. For example:

```rb
def insert_params
  super.tap {|p| p[:part] = "snippet,status"}
end
```

lib/yt/collections/live_broadcasts.rb
gmarziou: I must admit that I did not understand how this worked in other classes.
I looked at other collections and they seemed to use slightly different patterns.

Could you point me to a class that is a good example that I could study?

kangkyu: Thank you. @gmarziou

I was wondering if [this part](https://github.com/claudiob/yt/blob/master/lib/yt/collections/resources.rb#L31) could be used and simplify `insert` method above, how do you think? If you don't mind, could you please try something like following code, for me to learn more?

```rb
def insert_parts
  snippet = {keys: [:title, :description, :scheduled_start_time, :scheduled_end_time, :default_language], sanitize_brackets: true}
  status = {keys: [:privacy_status, :self_declared_made_for_kids]}
  {snippet: snippet, status: status}
end
```

gmarziou:
I think I applied your suggestions and tested, it works. 
I will squash if it's better now.

gmarziou:
Oops, I found that maybe the issue could come from the fact that LiveBroadcasts inherits from Base while it should inherit from Resources

Could you explain why some classes use Base super class (e.g. LiveCuepoints) while some like Playlists use Resources?

kangkyu:
I think Resources have title and description for them. (Yes, that's why) If you prefer `Base` directly, then I think you may simply remove `insert_parts` from your code. Either way. I don't have a strong opinion that `LiveBroadcast` should be one of `Resource` here.

gmarziou:
When I use `Resources` as parent class for `LiveBroadcasts`, it simplifies the code but it fails to build a valid snippet part
for `scheduled_start_time` and `scheduled_end_time` which get nullified by [Yt::Actions::Base#sanitize_brackets!](https://github.com/claudiob/yt/blob/master/lib/yt/actions/base.rb#L23)  because they are of type `TimeWithZone`.

```ruby
      def sanitize_brackets!(source)
        case source
          when String then source.gsub('<', '‹').gsub('>', '›')
          when Array then source.map{|string| sanitize_brackets! string}
          when Hash then source.each{|k,v| source[k] = sanitize_brackets! v}
        end
      end
```

I think that `sanitize_brackets!` should left as-is the types it can't handle, I propose this change:


```ruby
      def sanitize_brackets!(source)
        case source
          when String then source.gsub('<', '‹').gsub('>', '›')
          when Array then source.map{|string| sanitize_brackets! string}
          when Hash then source.each{|k,v| source[k] = sanitize_brackets! v}
          # Don't sanitize other types of objects
          else source
        end
      end
```

I tested it, it works fine for creating LiveBroadcasts.

kangkyu (approved):
Looks good to me

gmarziou:
Great, I just squashed the 3 commits on my branch.

